PR #24, opened 2021-07-16
Documentation 0.2.0 updated

anslpa:
Documentation 0.2.0 updated (is now using pyansys_data_theme). Have been published. This must be cherry-picked over release-0.2.0 feature branch.


docs/source/conf.py
```diff
@@ -115,29 +115,20 @@
...
-
-
-# Theme options are theme-specific and customize the look and feel of a theme
-# further.  For a list of options available for each theme, see the
-# documentation.
-#
 html_theme_options = {
-    "github_url": "https://github.com/pyansys/DPF-Core",
-    # "show_toc_level": 1,
-    # "show_prev_next": False
```
akaszynski, line 132: You might consider keeping `"show_prev_next": False` set.  Docs look better without it.

requirements_docs.txt
```diff
@@ -11,3 +11,4 @@ sphinx-notfound-page>=0.3.0
 sphinx-copybutton
 sphinx-gallery>=0.4.0
 pydata-sphinx-theme
```
akaszynski, line 13:
```suggestion
```
This can be removed since it's already used by the `pyansys-sphinx-theme`

requirements_docs.txt
```diff
@@ -11,3 +11,4 @@ sphinx-notfound-page>=0.3.0
...
+pyansys_sphinx_theme
```
akaszynski, line 14:
```suggestion
pyansys-sphinx-theme
```
It seems that it still installs, but it should be dashes instead of underscores.

akaszynski:
Please note that `Sphinx==4.1.1` has a bug where the logo won't be rendered correctly.  Hardcode `Sphinx==4.0.3` in `requirements_docs.txt` as a temporary fix until they fix this in the next release.

akaszynski (approved):
See suggestions.  Other than those, LGTM.

PR #54, opened 2021-11-30
Result data: on coordinates enabled, plotting updated

anslpa:
The result data can be requested on specific coordinates. 

With coordinates being a list of coordinates (e.g. [[-0.0195, 0.006, -0.0025], [-0.0192, 0.006, -0.0025], [-0.0189, 0.006, -0.0025], ...]), the following code can be used: 

```
from ansys.dpf import post
from ansys.dpf.post import examples
multishells = examples.multishells_rst
solution = post.load_solution(examples.multishells_rst)
path = post.create_path_on_coordinates(coordinates=coordinates)
displacement = solution.displacement(path=path)
displacement.vector.plot_contour(notebook=False)
```

image

**Next step**
Check documentation generated in CI/CD


ansys/dpf/post/dpf_path.py
```diff
@@ -66,35 +66,42 @@ def __init__(self, coordinates, scoping=None):
         ...     coord_copy[1] = coord_copy[0] + i * 0.001
         ...     coordinates.append(coord_copy)
         >>> dpf_path = post.DpfPath(coordinates=coordinates)
+
         """
-        self._coordinates = coordinates
+        # check inputs
         error_text = """coordinates attribute must
         be a list of list of int."""
         if not isinstance(coordinates, list):
```
cbellot000, line 74: @anslpa can't we support the Field here?
More generally speaking, excessively checking types in python is not a good practice (python idiomatic is [“it’s easier to ask for forgiveness than permission”](https://devblogs.microsoft.com/python/idiomatic-python-eafp-versus-lbyl/).)
Here for example, you will raise if coordinates is an numpy array. You should not make any check here, the line `self._field.data = coordinates` will raise a meaningful exception if necessary and field.data is the only method that knows what it supports

ansys/dpf/post/dpf_path.py
```diff
@@ -66,35 +66,42 @@ def __init__(self, coordinates, scoping=None):
...
             if len(coordinates) > 0 and not isinstance(coordinates[0], list):
                 raise ValueError(error_text)
-        self._scoping = scoping
-        if self._scoping is None:
-            self._scoping = Scoping(location=locations.nodal)
-            self._scoping.ids = list(range(1, len(coordinates) + 1))
+        _scoping = None
+        if scoping is None:
+            _scoping = Scoping(location=locations.nodal)
+            _scoping.ids = list(range(1, len(coordinates) + 1))
```
cbellot000, line 82: @anslpa is that not len(coordinates)/3 ?

anslpa: @cbellot000 It is len(coordinates) as far as I am expecting type(coordinates)=list[list[int]], but I can add some check in case I would have type(coordinates)=list[int]

ansys/dpf/post/dpf_path.py
anslpa: @cbellot000 I can add the possibility to define coordinates as a field, but I think this is needed to have the possibility to set coordinates as an array (as far as one of the goals of the ansys-dpf-post module is to have access to a post processing API thta does not require knowledges about ansys-dpf-core data model)

ansys/dpf/post/dpf_path.py
```diff
@@ -0,0 +1,75 @@
...
+from ansys.dpf.core import Field
+
+def create_path_on_coordinates(coordinates):
+    """
+    Create a dpf path object that can be used to request
+    results on a specific path of coordinates.
+
+    Parameters
+    ----------
+        coordinates : list[list[int]], Field, arrays
```
cbellot000, line 16: one tab too many here 
```suggestion
    coordinates : list[list[int]], Field, arrays
```

ansys/dpf/post/result_data.py
```diff
@@ -256,48 +297,56 @@ def plot_contour(
         [{'elshape': 1, 'time': 1}, {'elshape': 0, 'time': 1}]
```
cbellot000, line 297: common remark: get_all_label_spaces could be a good generic method for collections

cbellot000 (approved):

rlagha (approved):

PR #66, opened 2022-04-01
CI/CD fix related to OpenGL installation

anslpa:
Fix based on https://github.com/pyansys/pydpf-core/pull/196/files.


PProfizi:
Fix copied in #68

PR #84, opened 2022-06-30
Add Dependabot to the repository

PProfizi:
Adding dependabot to automatically check for new versions of required packages, as well as of github-actions


docs/source/conf.py
```diff
@@ -4,7 +4,7 @@
 
 import pyvista
 import numpy as np
-from pyansys_sphinx_theme import pyansys_logo_black
```
anslpa, line 7: What's the differences between pyansys_sphinx_theme and ansys_sphinx_theme  ?

.github/workflows/style.yml
```diff
@@ -1,7 +1,22 @@
...
+       - 'dependabot/*'
+  push:
+    tags:
+      - "*"
+    branches:
+      - master
+      - "release*"
+      - 'dependabot/*'
+
+concurrency:
```
anslpa, line 17: @PProfizi What's that for ?

docs/source/conf.py
PProfizi: The pyansys_sphinx_theme is the older, deprecated project, and ansys_sphinx_theme is the new one the pyansys libraries are supposed to use. https://github.com/ansys/ansys-sphinx-theme

requirements_docs.txt
```diff
@@ -1,13 +1,13 @@
...
-sphinx-autobuild
-sphinxcontrib-napoleon
-sphinxcontrib-websupport
-pytest-sphinx
-sphinx-notfound-page>=0.3.0
-sphinx-copybutton
-sphinx-gallery>=0.4.0
-pyansys-sphinx-theme
\ No newline at end of file
+pypandoc==1.8.1
```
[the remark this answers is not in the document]
PProfizi: @anslpa Yes, as explained by Alex [here](https://github.com/pyansys/pydpf-core/pull/275#issuecomment-1172365648), Dependabot's only job is to start PRs which will test any new version of a dependency in our requirement_* files.
This is to detect as soon as possible potential breaking changes, and act accordingly to solve the issue or update our setup.py.
So to do this, in our requirement files, all packages end-up pinned to the latest available version.

.github/workflows/style.yml
PProfizi: @anslpa Not sure which line you mean but:
The dependabot/* is to not run style check jobs on dependabot PRs (not needed)

The concurrency is to cancel previously running workflows for a PR when a new commit is made. This is to prevent clusters of jobs running for older commits than the latest. It is a choice, we can discuss tweaking that.

anslpa (approved):

PR #86, opened 2022-07-01
Remove _plot_chart docstring example AND restrain protobuf

PProfizi:
Removes the dosctring example for _plot_chart in result_data.py as it does not take the global PyVista env in account and freezes the pipelines.
Also cherrypicks the commit for restraining protobuf to 3.* since it also prevents the pipelines from running.


anslpa (approved):
Please do not forget to remove the ""protobuf<=3.20.1" statement for the incoming ansys-dpf-post releases (as well as for the ansys-dpf-core)

PR #116, opened 2022-08-08
Fix/memory leak and compatibility with core 0.5.2

PProfizi:


.github/workflows/release.yml
```diff
@@ -31,7 +31,7 @@ jobs:
       - uses: actions/checkout@v3
 
       - name: "Build Package"
-        uses: pyansys/pydpf-actions/build_package@v2.1
+        uses: pyansys/pydpf-actions/build_package@v2.2.dev1
```
cbellot000, line 34: @PProfizi why do we need to use 3 different versions of pydpf-actions here?

ansys/dpf/post/result_data.py
```diff
@@ -133,7 +133,10 @@ def num_fields(self):
     def get_data_at_field(self, field_index: int = 0):
         """Returns the data at the field with the specified index."""
         self._evaluate_result()
-        return self.result_fields_container[field_index].data
+        owning_field = self.result_fields_container[field_index]
```
cbellot000, line 136: @PProfizi let's maybe add a comment to remember the hack?

.github/workflows/release.yml
PProfizi: @cbellot000 Indeed, I was bumping all to 2.2.dev1 in core, was going to do the same here, now it is done.

cbellot000 (approved):

PR #148, opened 2022-09-02
Add a retro-compatibility check

PProfizi:


cbellot000 (approved):
In my opinion this can be done after the release

PR #164, opened 2022-09-23
Improve TypeHints

### LINKED ISSUE #160: Improve TypeHints print in the doc
### 📝 Description of the feature

Following [this pyansys guideline](https://dev.docs.pyansys.com/doc-style/docstrings.html#type-hints) would improve the layout of our API documentation by removing typehinting from the signature and instead putting it at the level of each parameter.
This also removes the need to write the type required at the dosctring level apparently.
See pyansys/pydpf-core#498.

### 💡 Steps for implementing the feature

https://dev.docs.pyansys.com/doc-style/docstrings.html#type-hints

### 🔗 Useful links and references

_No response_

GuillemBarroso:
Resolves #160. 

@PProfizi, note that the extensions are different in conf.py files from dpf-post and dpf-core. I was tempted to add all extensions from dpf-core to dpf-post, but in dpf-post we have 2 that are missing in dpf-core (notfound.extension and sphinx.ext.autosummary). I do not know if this is worth an issue and try to unify the extensions criterion. Thoughts?

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -47,14 +47,13 @@ extensions = [
     "sphinx.ext.autodoc",
     "sphinx.ext.autosummary",
     "sphinx.ext.intersphinx",
-    "sphinx.ext.autodoc.typehints",
     "sphinx.ext.napoleon",
+    "sphinx_autodoc_typehints",
     "numpydoc",
     "sphinx.ext.todo",
     "sphinx_copybutton",
     "sphinx_gallery.gen_gallery",
 ]
-autodoc_typehints = "description"
 
 # Intersphinx mapping
 intersphinx_mapping = {
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -44,12 +44,12 @@ release = version = __version__
 # ones.
 extensions = [
     "notfound.extension",
-    "sphinx.ext.autodoc.typehints",
-    "sphinx.ext.napoleon",
-    "numpydoc",
     "sphinx.ext.autodoc",
     "sphinx.ext.autosummary",
     "sphinx.ext.intersphinx",
+    "sphinx.ext.autodoc.typehints",
+    "sphinx.ext.napoleon",
+    "numpydoc",
     "sphinx.ext.todo",
     "sphinx_copybutton",
     "sphinx_gallery.gen_gallery",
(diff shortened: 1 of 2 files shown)


PProfizi:
> Resolves #160.
> 
> @PProfizi, note that the extensions are different in conf.py files from dpf-post and dpf-core. I was tempted to add all extensions from dpf-core to dpf-post, but in dpf-post we have 2 that are missing in dpf-core (notfound.extension and sphinx.ext.autosummary). I do not know if this is worth an issue and try to unify the extensions criterion. Thoughts?

@GuillemBarroso IMO the plan is to homogenize these type of files as muc has possible.
Some extensions might be useful only in one case for a specific reason, but I do not think it hurts to add them by default in the project it is not in. I am of the opinion of making this type of thing coherent.

GuillemBarroso:
@PProfizi I see! Could you please point me where I can find a good typehinting example in dpf-core so I can replicate it in dpf-post? 

Regarding the extensions, I was checking other repos (pymapdl and pyfluent) and they have a different extension that the one that is proposed in the [documentation](https://dev.docs.pyansys.com/doc-style/docstrings.html#type-hints). The documentation proposes "sphinx.ext.autodoc.typehints", but pymapdl and pyfluent have "sphinx_autodoc_typehints", see [doc on sphinx_autodoc_typehints](https://pypi.org/project/sphinx-autodoc-typehints/). However, I have been trying to include "sphinx_autodoc_typehints" and the documentation file keeps being of size 7KB.

PProfizi:
> @PProfizi I see! Could you please point me where I can find a good typehinting example in dpf-core so I can replicate it in dpf-post?
> 
> Regarding the extensions, I was checking other repos (pymapdl and pyfluent) and they have a different extension that the one that is proposed in the [documentation](https://dev.docs.pyansys.com/doc-style/docstrings.html#type-hints). The documentation proposes "sphinx.ext.autodoc.typehints", but pymapdl and pyfluent have "sphinx_autodoc_typehints", see [doc on sphinx_autodoc_typehints](https://pypi.org/project/sphinx-autodoc-typehints/). However, I have been trying to include "sphinx_autodoc_typehints" and the documentation file keeps being of size 7KB.

About the documentation failing, for later reference, the documentation generation started failing on 9/24 due to a time-out during sphinx generation. trying to fix this right now. So your problem might not be due to the extension.

PProfizi (approved):
@GuillemBarroso LGTM, still the remark about there not being a lot of typehints in the Post doc to begin with. We'll see how this goes with usage.

PR #225, opened 2022-12-20
Add Simulation

PProfizi:
Refactors Solution to Simulation.
Adds load_simulation command.

One drawback is that users would have to change the load_solution command in their scripts to load_simulation.
But then again, with all the changes we will make, they will have to change their scripts anyway and this way it would make a clear difference as to which version they are using.

This also wraps FieldsContainer with DataObject instead of inheriting, solving a memory issue and simplifying manipulation of the fields_container. 

This then fixes DataObject.as_array and as_dataframe so that docstring tests pass.


src/ansys/dpf/post/data_object.py
```diff
@@ -66,38 +67,61 @@ def as_data_frame(self, columns=None, **kwargs):
         >>> import pandas as pd
         >>> from ansys.dpf import post
         >>> from ansys.dpf.post import examples
-        >>> solution = post.load_solution(examples.multishells_rst, legacy=False)
-        >>> #stress = solution.stress()
-        >>> #stress.xx.plot_contour(show_edges=False)
+        >>> solution = post.load_simulation(examples.multishells_rst)
```
GuillemBarroso, line 70:
```suggestion
        >>> simulation = post.load_simulation(examples.multishells_rst)
```
I'd go for simulation since we use the ``.load_simulation()`` method. I prefer simulation over solution.

src/ansys/dpf/post/data_object.py
```diff
@@ -66,38 +67,61 @@ def as_data_frame(self, columns=None, **kwargs):
...
+        >>> # Export the displacements vector field at step 1 as a DataFrame
+        >>> displacement = solution.displacement(steps=[1], nodes=[1, 2, 3])
```
GuillemBarroso, line 72:
```suggestion
        >>> displacement = simulation.displacement(steps=[1], nodes=[1, 2, 3])
```

src/ansys/dpf/post/data_object.py
```diff
@@ -66,38 +67,61 @@ def as_data_frame(self, columns=None, **kwargs):
...
-        #     columns.appends(arg)
 
-        # #load data into a DataFrame object:
-        # data_frame = pd.DataFrame(self.field(0), columns)
-        # transposed = data_frame.transpose()
         if not columns:
             columns = self._columns
-        return pd.DataFrame(
+        # The wrapped FieldsContainer should only contain one Field
+        if len(self) > 1:
```
GuillemBarroso, line 91: I am going to add another point in the metaissue under the ``DataObject`` category about extending the as_data_frame to more than one field.

src/ansys/dpf/post/data_object.py
```diff
@@ -66,38 +67,61 @@ def as_data_frame(self, columns=None, **kwargs):
...
+        Returns
+        -------
+        class:`pandas.core.frame.DataFrame`
+
+        Examples
+        --------
+        >>> import pandas as pd
+        >>> from ansys.dpf import post
+        >>> from ansys.dpf.post import examples
+        >>> solution = post.load_simulation(examples.multishells_rst)
```
GuillemBarroso, line 114:
```suggestion
        >>> simulation = post.load_simulation(examples.multishells_rst)
```

src/ansys/dpf/post/data_object.py
```diff
@@ -66,38 +67,61 @@ def as_data_frame(self, columns=None, **kwargs):
...
+        >>> # Export the displacements vector field at step 1 as a DataFrame
+        >>> displacement = solution.displacement(steps=[1], nodes=[1, 2, 3])
```
GuillemBarroso, line 116:
```suggestion
        >>> displacement = simulation.displacement(steps=[1], nodes=[1, 2, 3])
```

src/ansys/dpf/post/data_object.py
```diff
@@ -66,38 +67,61 @@ def as_data_frame(self, columns=None, **kwargs):
...
+        >>> arr = displacement.as_array()
+        >>> print(arr)
+        [[  0.39831985 -13.79737819  -0.16376683]
+         [  0.23311378 -13.79765179  -0.15318972]
+         [  0.36754181 -13.80815135  -0.16373945]]
+
+        """
+        return self._fc[-1].data
```
GuillemBarroso, line 124: I think returning the last field should be fine for most cases (static, transient mechanical, fluids, etc). But for modal analysis we should return the 1st mode. I am not sure if that is the only case that is different tho.

GuillemBarroso (approved):
Great, thanks Paul! Just left a couple comments. Everything else LGTM.

src/ansys/dpf/post/data_object.py
PProfizi: @GuillemBarroso Thx for the catch!

src/ansys/dpf/post/data_object.py
PProfizi: > I think returning the last field should be fine for most cases (static, transient mechanical, fluids, etc). But for modal analysis we should return the 1st mode. I am not sure if that is the only case that is different tho.

@GuillemBarroso You are right, I'll have to change that behavior once I implement the "ModalSimulation" type or something like that. The DataObject should thus have an idea of the type of simulation it is coming from.

GuillemBarroso: Great!

PR #257, opened 2023-01-10
Add pseudocode for Mechanical Simulations

PProfizi:
Keeps track of pseudo-code for each Mechanical simulation types:
- Static
- Transient
- Modal
- Harmonic


temp/harmonic_analysis.py
```diff
@@ -0,0 +1,98 @@
...
+
+# Extract nodal XY stresses for elements 1, 2 and 3 at frequency 1
+stress_XY = harmonic_simulation.elemental_stress(
+    component="XY", elements=[1, 2, 3], steps=[1]
+)
+print(stress_XY)
+
+# Extract first principal nodal stress for a named (elemental or nodal) selection at all frequencies
+stress_S1 = harmonic_simulation.nodal_stress(
+    component="S1", named_selection=named_selections[0]
```
MaxJPRey, line 77: @PProfizi Should `nodal_stress` be renamed by `nodal_principal_stress` such as for static.
```suggestion
stress_S1 = harmonic_simulation.nodal_principal_stress(
    component=1, named_selection=named_selections[0]
```

MaxJPRey: Same question goes for `modal_simulation` below.

temp/static_analysis.md
```diff
@@ -0,0 +1,549 @@
...
+- named selection: {named_selection[0]}
+```
+
+### Get the minimum, maximum, mean of the DataObject 
+Use DPF operators for efficiency.
+
+Across all dimensions present in the specific DataObject, unless specified as argument
+Take inspiration from the "axis" argument for ``numpy.ndarray.amax``.
+Do we want a ``nanmax``? Which propagates NaN values? (Whether we have actual NaNs or not)
+These would return DataObjects or a scalar.
```
germa89, line 451:
```suggestion
Do we want a ``nanmax``? Which propagates NaN values? (Whether we have actual NaNs or not)
These would return DataObjects or a scalar.
```
I don't think we need that... if you return the data as an array, you can filter it.

temp/static_analysis.md
```diff
@@ -0,0 +1,549 @@
...
+```
+
+Plot a one-step DataObject - the "step" argument is not read
+```pycon
+>>> stress_S1_step_1.plot()
+```
+
+Plot principal stresses
+
+TODO: Not really sure how this is different from plotting results
```
germa89, line 493: The filtering is more granular. But I don't think it add much more.

temp/static_analysis.md
```diff
@@ -0,0 +1,549 @@
...
+>>> max_stress_S1_node_1.animate_graph(axis="steps", x="ID")
+```
+
+### Interpolate data from one mesh to another
+
+Starts from the mesh support of the DataObject
+
+Performs interpolation for each step and for each component present by default (for each field)
+```pycon
+>>> stress_S1_on_another_mesh = stress_S1.interpolate(mesh=another_mesh)
```
germa89, line 534: This is the one in my opinion.

temp/static_analysis.py
```diff
@@ -0,0 +1,340 @@
...
+# Create and use geometry
+# Creation of geometry is for now in PyDPF-Core.
+# Do we expose those in Post in some way? Simply by importing them in a specific Post module?
+from ansys.dpf.core.geometry import Line
+
+line = Line(coordinates=[[1.0, 0.0, 1.0], [5.0, 0.0, 1.0]])
+static_simulation.geometries.append(line)
+# static_simulation.plot(mesh=True, geometry=True, boundary_conditions=True)
+
+
```
germa89, line 64: I'm not sure this is needed/supported at this point.

Adding a geometry might seem we can change the mesh with the new geometry and solve?

I rather keep the geometry as "only read" for the moment.

temp/static_analysis.py
```diff
@@ -0,0 +1,340 @@
...
+# TODO
+
+# Set a selection as default for a simulation for further result extractions
+static_simulation.activate_selection(selection=selection)
+
+# Remove a selection as default for a simulation for further result extractions
+static_simulation.deactivate_selection()
+
+# Apply a SpatialSelection to a Simulation to get resulting list of entities IDs (nodes or elements)
+list_of_IDs = selection.spatial_selection.apply_to(static_simulation)
```
germa89, line 143: Same as my previous comment.

germa89 (changes requested):
Several changes have been proposed, but overall, I think this is the way to go.

temp/harmonic_analysis.py
```diff
@@ -0,0 +1,96 @@
...
+from ansys.dpf.post.selection import Selection
+
+# Provide Enums for available physics_types and analysis_types
+# from ansys.dpf.post import physics_types, analysis_types
+
+
+harmonic_simulation = dpf.load_simulation(examples.complex_rst)
+# harmonic_simulation = dpf.load_simulation(examples.complex_rst,
+#                                           physics_type=physics_types.mechanical,
+#                                           analysis_type=analysis_types.harmonic)
```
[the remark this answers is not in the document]
PProfizi: Hi @germa89, as I mentioned in the other PR, the arguments for this API have been changed. I updated the code in this PR to reflect that.

temp/harmonic_analysis.py
```diff
@@ -0,0 +1,96 @@
...
+# Explore the simulation metadata
+
+# Print information about the simulation
+print(harmonic_simulation)
+
+# Print the mesh
+print(harmonic_simulation.mesh)
+
+# Print the list of constructed geometries
+print(harmonic_simulation.geometries)
```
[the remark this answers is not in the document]
PProfizi: I personnally expect this property to return a list of existing geometry objects, hence the plural for now.

temp/harmonic_analysis.py
```diff
@@ -0,0 +1,96 @@
...
+print(harmonic_simulation.rpms)
+
+# Print the list of available results
+print(harmonic_simulation.results)
+
+# General plot of the simulation object with, by default:
+# - the mesh, not deformed
+# - the geometry
+# - the boundary conditions
+harmonic_simulation.plot(mesh=True, geometry=True, boundary_conditions=True)
```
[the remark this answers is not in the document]
PProfizi: @germa89, yes we should. This PR is to merge in the main refactor branch a first draft of the pseudocode.
I have not treated the geometry or boundary_conditions API yet.
There is however already the mesh.plot and the plot for each geometry object done by Guillem.
They are specified in the Mesh and the Geometry sections respectively in the static pseudocode.

temp/harmonic_analysis.py
```diff
@@ -0,0 +1,96 @@
...
+
+# -----------------------------------------------------------------------------------------
+# Apply a selection
+selection = Selection()
+selection.select_nodes(nodes=[1, 2, 3])
+selection.select_time_freq_indices(time_freq_indices=[0, 1])
+harmonic_simulation.activate_selection(selection_object=selection)
```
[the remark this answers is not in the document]
MaxJPRey: I agree @germa89 
In that case, should we have properties `nodes`, `time_freq_indices` available on `Selection` objects?
In that case, the methods `select_...()` can definitely be removed.
Maybe it is already the case, but just in case.

temp/harmonic_analysis.py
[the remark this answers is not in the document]
cbellot000: I disagree for this one (we accept can both for sure) but "bringing in more objects" is a dev problem and not a user one whereas having to know strings and potentially making typos is a user problem

temp/harmonic_analysis.py
```diff
@@ -0,0 +1,96 @@
...
+print(harmonic_simulation.boundary_conditions)
+
+# Print the list of available named selections
+named_selections = harmonic_simulation.named_selections
+print(named_selections)
+
+# Print available frequencies, phase angles, and RPMs
+print(harmonic_simulation.time_freq_support)
+print(harmonic_simulation.frequencies)
+print(harmonic_simulation.phase_angles)
```
cbellot000, line 38: what is phase_angles?

temp/harmonic_analysis.py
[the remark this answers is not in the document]
cbellot000: @PProfizi I'm not sure to understand, how do you plot boundary conditions? will you plot all geometries added to the simulation?

temp/harmonic_analysis.py
```diff
@@ -0,0 +1,96 @@
...
+print(stress_XY)
+
+# Extract first principal nodal stress for a named (elemental or nodal) selection at all frequencies
+stress_S1 = harmonic_simulation.nodal_principal_stress(
+    components=["1"], named_selection=named_selections[0]
+)
+print(stress_S1)
+
+# Extract equivalent elemental strain for a selection at set 1
+strain_eqv = harmonic_simulation.elemental_eqv_strain(selection=selection, set_ids=[1])
```
cbellot000, line 82: I would reverse the names: elemental_strain_eqv, nodal_stress_principal, what do you think?

temp/harmonic_analysis.py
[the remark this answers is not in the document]
cbellot000: @PProfizi could you also show how you add geometries? 
Should we rename geometries to constructed geometries? geometry makes me think that we are getting CAD info...

temp/static_analysis.md
```diff
@@ -0,0 +1,549 @@
...
+[1]
+>>> print(static_simulation.nsteps)
+1
+```
+
+Print the list of available results
+```pycon
+>>> print(static_simulation.results)
+TBD
+```
```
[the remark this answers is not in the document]
cbellot000: @germa89 it appears that Paul meant "available results" by results

temp/static_analysis.md
```diff
@@ -0,0 +1,549 @@
...
+>>> load = Load(data=data_object)
+>>> print(load)
+TBD
+```
+
+### Map loads from one mesh to another 
+(which is the same as DataObject.interpolate?)
+```pycon
+>>> load.interpolate(mesh=another_mesh)
+```
```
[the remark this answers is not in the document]
cbellot000: I agree

temp/static_analysis.md
```diff
@@ -0,0 +1,549 @@
...
+
+### Create a selection
+```pycon
+>>> selection = Selection()
+>>> selection.select_nodes(nodes=[1, 2, 3])
+>>> selection.select_elements(elements=[1, 2, 3])
+>>> selection.select_time_freq_indices(time_freq_indices=[1])  # Rework?
+>>> # selection.select_steps(steps=[1])
+>>> # selection.select_geometry(geometry=Line())
+>>> selection.select_named_selection(named_selection=named_selections[0])
```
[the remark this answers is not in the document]
cbellot000: @germa89 I don't mind the reduncy as long as we have other options. What we decided in previous meetings was to work with factories and make people do:
```suggestion
>>> selection = tools.create_selection(nodes=[1, 2, 3], elements=[1, 2, 3], steps=[1])
```
which is already merged BTW

temp/static_analysis.md
```diff
@@ -0,0 +1,549 @@
...
+```pycon
+>>> load_2 = mesh.interpolate(load)
+```
+
+## Extract specific results
+
+Arguments:
+- components: list of components amongst
+  - [1, 2, 3, 4, 5, 6] equivalent to ["X", "Y", "Z", "XY', "YZ", "XZ"], or "N" for the norm
+  - [1, 2, 3] for S1, S2, S3, E1, E2, or E3 when requesting principal stresses/strains
```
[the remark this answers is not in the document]
PProfizi: @germa89 
I am not sure I understand which 1, 2, 3 you refer to when you say it would be confusing?

Also, the way I am designing it now, "all" would be equivalent to not specifying components.
As for "norm", I am not sure how to treat it.
It could be indeed:
- an additional component
- an additional parameter, which means we take the norm of the components asked?

temp/static_analysis.md
```diff
@@ -0,0 +1,549 @@
...
+These would return DataObjects or a scalar.
+```pycon
+>>> global_max_stress_S1 = stress_S1.max()
+```
+```pycon
+>>> max_stress_S1_step_1 = stress_S1_step_1.max()
+```
+or
+```pycon
+max_stress_S1_step_1 = stress_S1.max(steps=[1])
```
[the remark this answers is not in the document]
PProfizi: @germa89 there is a difference though between the way I wrote it and the way you did:
- In my case, extraction was defined before, and you get a dataset across many dimensions
- In your case you first extract to only the dimension of interest for your `max` operation. So you would end-up extracting again just to have data specific to this operation.

Your iterate method shows that we indeed need a way to tell our methods (max min or any other) on which dimensions to work. I took inspiration from the `numpy.ndarray.max` method which does have the `axis` argument to know on which dimension to work. Many of their methods have ways to specify the wanted behavior for the operation.
This is something we need to discuss, as you said.

temp/static_analysis.py
PProfizi: The goal of the constructed geometry is to allow for creation of supports for operations such as mapping and result evaluation on sections... etc.
It is part of what we will need the user to be able to add and interact with. Right now it is just a list of "geometry" objects, but it could be more.
To me, `geometry`, is different from `mesh` and as long as we document it properly it should be ok.
Also we do not have the possibility to generate input files from DPF.

But this definitely needs to be better defined in terms of what is needed.

temp/static_analysis.py
PProfizi: I am not sure I see which previous comment you refer to

temp/static_analysis.md
PProfizi: Not sure I understand what you mean.

temp/static_analysis.md
```diff
@@ -0,0 +1,549 @@
...
+>>> max_stress_S1_node_1.graph(x="steps", names="ID")
+```
+or as proposed in the GitHub discussion #213
+```pycon
+>>> max_stress_S1_node_1.plot(graph=True, individual_figures=False)
+```
+
+Plot waterfall/stagger diagrams -> a waterfall diagram is different from a waterfall plot!
+
+Plot waterfall plots
```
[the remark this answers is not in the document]
PProfizi: I think the difference is that the way you wrote it, the filtering is done at the extraction step, which is of course a possibility but a specific extraction should not be required before every operation.
We will need the possibility to work on only part of a DataObject via filtering after extraction.
Your proposed `.iterate()` method is one way of solving part of this issue, yet I think we will need `.plot` methods for our objects, with DataObject-specific arguments.
In your example, the data is filtered and treated so it can be used by the matplotlib plot function. This is one way of doing it but IMHO our goal is to provide helpers so that users can do all this using our APIs.

temp/harmonic_analysis.py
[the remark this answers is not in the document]
PProfizi: @germa89 
Do you mean something like this?
```py
selection.activate(simulation=Simulation)
```

The problem with `selection.deactivate()` is that a Selection is not linked to a Simulation. It is applied and evaluated. The simulation knows which selection is applied, but a selection does not know which simulation instances it has been applied to. IMHO this mapping is not necessary.
I can understand the use-case for the `selection.activate(simulation=Simulation)` command as a helper, but deactivating a selection should, I think, be done at the simulation level.

We do have also for now:
```py
SpatialSelection.apply_to(simulation=Simulation)
TimeFreqSelection.apply_to(simulation=Simulation)
```
which each return a list of IDs of nodes/elements or steps/timesteps/modes respectively.

temp/static_analysis.md
PProfizi: @germa89 I am not sure I understand what is specific in the case of plotting principal stresses.

IMO since this section is about plotting 3D contours, only one principal stress can be plotted at a time. Was the point to mention that the `DataObject.plot()` method should accept a component input argument?
With the refactoring of DataObject as DPFDataframe, I think this kind of selection will be made via columns labels. For example:
```py
df = simulation.principal_stress_elemental()
# df is now a DPFDataframe with three named 'columns': 'S1', 'S2', 'S3'
df.plot(column='S2')
```
and it would take the "first" column by default, here 'S1'.

Although, `pandas.Dataframe.plot` does plot the given data as a curve, and not as a 3D contour since they do not have the notion of mesh support for the data. Maybe we should rethink the naming.

temp/harmonic_analysis.py
PProfizi: @cbellot000 that would be the list of values for the sweeping phase. I think I took that naming from somewhere...

temp/modal_analysis.md
```diff
@@ -0,0 +1,57 @@
...
+```pycon
+def result(
+        self,
+        components: Union[str, List[str], int, List[int], None] = None,
+        norm: bool = False,
+        selection: Union[Selection, None] = None,
+        frequencies: Union[float, List[float], None] = None,
+        set_ids: Union[int, List[int], None] = None,
+        load_steps: Union[int, List[int], 
+            Tuple[Union[int, List[int], Union[int, List[int]], None] = None,
```
PProfizi, line 18:
```suggestion
        modes: Union[int, List[int], None] = None,
```

PProfizi:
Close as pseudocode of code merged in the meantime

PR #258, opened 2023-01-10
Add different Mechanical simulations

PProfizi:
In coherence with the proposed pseudocode, this PR adds the different mechanical simulation classes:
- StaticMechanicalSimulation
- TransientMechanicalSimulation
- ModalMechanicalSimulation
- HarmonicMechanicalSimulation

A first few additions are made to enable future work.

### DIFF
--- a/src/ansys/dpf/post/post_utility.py
+++ b/src/ansys/dpf/post/post_utility.py
@@ -210,7 +210,9 @@ def load_simulation(
         else:
             raise ValueError(f"Unknown physics type '{physics_type}.")
 
-    if issubclass(simulation_type, Simulation):
+    if simulation_type in [
+        getattr(AvailableSimulationTypes, x) for x in vars(AvailableSimulationTypes)
+    ]:
         return simulation_type(data_sources, _model)
     else:
         raise ValueError(
--- a/src/ansys/dpf/post/post_utility.py
+++ b/src/ansys/dpf/post/post_utility.py
@@ -6,12 +6,18 @@ import warnings
 
 from ansys.dpf.core.model import Model
 
-from ansys.dpf.post.common import _AnalysisType, _AvailableKeywords, _PhysicsType
+from ansys.dpf.post.common import (
+    AvailableSimulationTypes,
+    _AnalysisType,
+    _AvailableKeywords,
+    _PhysicsType,
+)
 from ansys.dpf.post.harmonic_analysis import HarmonicAnalysisSolution
 from ansys.dpf.post.modal_analysis import ModalAnalysisSolution
 from ansys.dpf.post.simulation import (
     HarmonicMechanicalSimulation,
     ModalMechanicalSimulation,
+    Simulation,
     StaticMechanicalSimulation,
     TransientMechanicalSimulation,
 )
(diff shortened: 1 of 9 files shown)


src/ansys/dpf/post/simulation.py
```diff
@@ -238,87 +393,17 @@ def displacement(
...
+class TransientMechanicalSimulation(MechanicalSimulation):
+    """Provides methods for mechanical transient simulations."""
+
+
+class ModalMechanicalSimulation(MechanicalSimulation):
+    """Provides methods for mechanical modal simulations."""
+
+
+class HarmonicMechanicalSimulation(MechanicalSimulation):
+    """Provides methods for mechanical harmonic simulations."""
```
germa89, line 409:
```suggestion
class StaticMechanicalSimulation(MechanicalSimulation):
    """Provides methods for mechanical static simulations."""
    pass


class TransientMechanicalSimulation(MechanicalSimulation):
    """Provides methods for mechanical transient simulations."""
    pass


class ModalMechanicalSimulation(MechanicalSimulation):
    """Provides methods for mechanical modal simulations."""
    pass


class HarmonicMechanicalSimulation(MechanicalSimulation):
    """Provides methods for mechanical harmonic simulations."""
    pass
```

src/ansys/dpf/post/post_utility.py
```diff
@@ -167,28 +182,43 @@ def load_simulation(
...
+        ):
+            if analysis_type == _AnalysisType.static:
+                simulation_type = StaticMechanicalSimulation
+            elif analysis_type == _AnalysisType.modal:
+                simulation_type = ModalMechanicalSimulation
+            elif analysis_type == _AnalysisType.harmonic:
+                simulation_type = HarmonicMechanicalSimulation
+            elif analysis_type == _AnalysisType.transient:
+                simulation_type = TransientMechanicalSimulation
+            else:
```
germa89, line 206: We should also allow comparison like:
```py
>>> analysis_type = "harmonic"
>>> analysis_type == _AnalysisType.static
False
```

germa89: This could be done either overloading the "==" operator for the a new class AnalysisType:
```py
class AnalysisType:
     def __eq__(self, other):
           ....
_AnalysisType.static = AnalysisType("static")
assert analysis_type == _AnalysisType.static
assert "static" == _AnalysisType.static
```
 or just write the text comparison in the code:
```py
if analysis_type == _AnalysisType.static or analysis_type.lower() == "static":
    simulation_type = StaticMechanicalSimulation
elif analysis_type == _AnalysisType.modal or analysis_type.lower() == "modal":
    simulation_type = ModalMechanicalSimulation
elif analysis_type == _AnalysisType.harmonic or analysis_type.lower() == "harmonic":
    simulation_type = HarmonicMechanicalSimulation
elif analysis_type == _AnalysisType.transient or analysis_type.lower() == "transient":
    simulation_type = TransientMechanicalSimulation

else:
    raise ValueError(f"The analysis_type '{analysis_type}' is not accepted.")
```

germa89 (approved):

src/ansys/dpf/post/post_utility.py
PProfizi: @germa89 actually the thing is we wanted to not expose _AnalysisType or _PhysicsType which come from pydpf-core, and instead give pre-canned strings to the user to define the exact combination they want (coresponging to available SimulationTypes).
We could accept string values, but these would not simply be "static" or "transient", these would be "static_mechanical" or "transient_mechanical". I feel like these give too many possibilities for errors and parsing them would not be robust.
What do you think?

PR #304, opened 2023-03-03
Fix iselect, update DataFrame.axes, Naming convention

PProfizi:
- DataFrame.axes now returns [DataFrame.index, DataFrame.columns], the two MultiIndex objects, like in Pandas.
- Fix bug for iselect when selecting on a label with a list of indices.
- Make naming coherent between result query API arguments and DataFrame indexes default names. Increase default width of columns for printed DataFrame.


cbellot000 (approved):
that's great, thanks!

PR #306, opened 2023-03-03
Add Dataframe.array

PProfizi:


src/ansys/dpf/post/dataframe.py
```diff
@@ -9,6 +9,7 @@
 import ansys.dpf.core as dpf
 from ansys.dpf.core.common import shell_layers
 from ansys.dpf.core.dpf_array import DPFArray
+import numpy as np
```
cbellot000, line 12: @PProfizi do we need to import numpy here?

cbellot000 (approved):

PR #328, opened 2023-03-10
Improvements to pyproject.toml - contact email

RobPasMue:
At the https://github.com/pyansys/pyansys-dev package we are experiencing some issues with a recent modification to your pyproject.toml file...

We have the feeling that it may be caused by the fact that you may be calling the entire repository as a dependency, while we are targeting a specific branch (``master``). In any case, if you want to leave it as it was, it would also be fine. It'd be more accurrate to just target the ``master`` branch on your side for this use case.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -11,13 +11,13 @@ readme = "README.md"
 requires-python = ">=3.7.*,<4.0"
 license = {file = "LICENSE"}
 authors = [
-    {name = "Ansys, Inc.", email = "pyansys.support@ansys.com"},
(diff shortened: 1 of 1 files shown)


RobPasMue:
Nah, probably not needed... Let me revert them. But the contact email should be changed in any case :)

PProfizi:
> Nah, probably not needed... Let me revert them. But the contact email should be changed in any case :)

Thank you for the changes!
This direct dependency to the git repo for ansys-dpf-core dates back to the addition of the pyproject.toml. I think it clones the repo and checks-out master, but since master is default, I do not think it downloads more than needed.

PProfizi (approved):

PR #330, opened 2023-03-10
Add units to ResultsIndex

PProfizi:
- [x] Store unit of each result in `ResultsIndex`.
- [x] Have them appear in the string representation of `ResultsIndex`, and thus of `DataFrame`.

We should also think of manipulating UnitSystem objects from PyDPF-Core -> at the Simulation level, to switch UnitSystem.
- [ ] Expose a ResultsIndex.units(List[str]) setter? Exposed at the DataFrame level to have DataFrame.units() setter, checking if units are the same and if not, tries to apply a `math.unit_convert_fc`.

### DIFF
--- a/src/ansys/dpf/post/index.py
+++ b/src/ansys/dpf/post/index.py
@@ -168,14 +168,21 @@ class ResultsIndex(Index):
         units: Union[List[Union[str, None]], None] = None,
     ):
         """Initiate this class."""
+        # Initialize results labels
         result_values = values
-        if units is not None:
+        if units is None:
+            # If no units, initialize as list of None
+            units = [None] * len(values)
+        else:
+            # Truncate units if longer than values
+            if len(units) > len(values):
+                units = units[: len(values)]
+            # Update result labels with unit when necessary
             for i, unit in enumerate(units):
                 result_values[i] += f" ({unit})" if unit is not None else ""
+            # If units was shorter than values, extend with Nones
             if len(units) < len(values):
                 units.extend([None] * (len(values) - len(units)))
-        else:
-            units = [None] * len(values)
         self._units = units
         super().__init__(name=ref_labels.results, values=result_values, scoping=None)
 
--- a/src/ansys/dpf/post/index.py
+++ b/src/ansys/dpf/post/index.py
@@ -165,9 +165,24 @@ class ResultsIndex(Index):
     def __init__(
         self,
         values: List[str],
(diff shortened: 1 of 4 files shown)


tests/test_dataframe.py
```diff
@@ -139,9 +139,15 @@ def test_dataframe_animate(transient_rst):
...
+        "index=MultiIndex<["
+        "MeshIndex<name=\"node_ids\", dtype=<class 'int'>>, "
+        "Index<name=\"components\", dtype=<class 'str'>>"
+        "]>, "
+        "columns=MultiIndex<["
+        "ResultIndex<['U (m)']>, "
+        "SetIndex<values=[1]>"
+        "]>>"
     )
     assert repr(df) == ref
```
germa89, line 152: You might want to use this pytest plugin to automatize "golden" tests

https://github.com/oprypin/pytest-golden#pytest-golden

Feel free to ping me about this.

germa89 (approved):
LGTM! :)

src/ansys/dpf/post/index.py
```diff
@@ -165,9 +165,24 @@ class ResultsIndex(Index):
...
         """Initiate this class."""
-        super().__init__(name=ref_labels.results, values=values, scoping=None)
+        result_values = values
+        if units is not None:
+            for i, unit in enumerate(units):
+                result_values[i] += f" ({unit})" if unit is not None else ""
+            if len(units) < len(values):
+                units.extend([None] * (len(values) - len(units)))
+        else:
+            units = [None] * len(values)
```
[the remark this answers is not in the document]
PProfizi: > Probably we want to avoid writing `"()"` if `unit` is `None`. :)
> 
> And personally, I read better the `if` before the `for`. But this is a nit pick :).

Hi @germa89! Thx for the review! Just to nit pick too, the original line 174 does actually write `""` in `unit` is `None` ;)

Also, the `if` after the `for` is because the operation of _adding the unit to the string rep_ if not needed for `None` values of units, which is all we are adding in the `if`. 
Thus having the `if` before the `for` might be a bit more readable, but is adds unnecessary operations. What I will do is add comments so it is more readable because you are right, it is not really readable.

tests/test_dataframe.py
PProfizi: oh nice, thx @germa89, I did not know this plugin!

PProfizi: Do you know if it works well for numerical values?

PR #333, opened 2023-03-13
Add DataFrame.min and DataFrame.max over 'set_ids' axis or MeshIndex

PProfizi:


tests/test_dataframe.py
```diff
@@ -258,3 +259,38 @@ def test_dataframe_array_raise(transient_rst):
...
     ):
         _ = df.array
+
+
+def test_dataframe_min_max():
+    simulation = post.TransientMechanicalSimulation(examples.download_crankshaft())
+    df = simulation.displacement(all_sets=True)
+    # Over the mesh entities
+    min_over_mesh = [[-0.00074732, -0.00040138, -0.00021555]]
+    assert np.all(np.isclose(df.min()._fc[0].data.tolist(), min_over_mesh))
```
MaxJPRey, line 269: Is it expected to expose and use a protected attribute `_fc`?

src/ansys/dpf/post/dataframe.py
```diff
@@ -833,3 +835,117 @@ def animate(
...
+
+    def min(self, axis: Union[int, str, None] = 0) -> Union[DataFrame, float]:
+        """Return the minimum value over the requested axis.
+
+        Parameters
+        ----------
+        axis:
+            Axis to perform minimum across.
+            Defaults to the MeshIndex (0).
+            Can also be the SetIndex (1).
```
MaxJPRey, line 847: Maybe here we could add additional info just to remind what are the `MeshIndex` and `SetIndex` for new users.
I don't remember if it is done somewhere else.

tests/test_dataframe.py
PProfizi: @MaxJPRey not to the user. We do have `._core_object` property though which we will implement for all classes wrapping a Core entity to give easy access in case it is needed. 
Here in the tests this allows me to compare the result of the max operation with doing the same thing directly on the data held by the underlying fields container.

examples/01-Detailed-Examples/06-compute-statistics.py
```diff
@@ -0,0 +1,57 @@
+"""
+.. _ref_compute_statistics_example:
+
+Compute statistics of a DataFrame
```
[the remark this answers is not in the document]
PProfizi: @cbellot000 the idea was to create this example now with min and max, and add to it as new features like mean are done

examples/01-Detailed-Examples/06-compute-statistics.py
```diff
@@ -0,0 +1,57 @@
...
+=================================
+In this example, transient mechanical displacement data is used
+to show how to compute statistical values such as the min, max or mean.
```
JennaPaikowsky:
```suggestion
to show how to compute statistical values such as the min, max, or mean.
```

examples/01-Detailed-Examples/06-compute-statistics.py
```diff
@@ -0,0 +1,57 @@
...
+"""
+
+###############################################################################
+# Perform required imports
+# ------------------------
+# Perform required imports.
```
JennaPaikowsky, line 14: Don't think we need this line twice for such a short section.

src/ansys/dpf/post/dataframe.py
```diff
@@ -833,3 +835,121 @@ def animate(
...
+        """Return the maximum value over the requested axis.
+
+        Parameters
+        ----------
+        axis:
+            Axis to perform maximum across.
+            Defaults to the MeshIndex (0), the row index containing mesh entity IDs.
+            This computes the maximum across the mesh for each set.
+            Can also be the SetIndex (1), the column index containing set (time/freq) IDs.
+            This computes the maximum across sets (time/freq) for each mesh entity.
```
JennaPaikowsky:
```suggestion
            Can also be the SetIndex (1), the column index containing set (time/frequency) IDs.
            This computes the maximum across sets (time/frequency) for each mesh entity.
```

examples/01-Detailed-Examples/06-compute-statistics.py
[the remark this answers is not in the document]
cbellot000: I understand but I think statistics is bit misleading. Personally, when I'm looking to compute a min max over time (which is a recuring need), I don't search for "statistics"

JennaPaikowsky (approved):

examples/01-Detailed-Examples/06-compute-statistics.py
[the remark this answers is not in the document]
PProfizi: @cbellot000 I understand. I am renaming it, and we will see how to reorganize once there are more features.

PR #354, opened 2023-04-03
Update compatibility.rst for DPF Server 2023.2.pre1

PProfizi:


docs/source/getting_started/compatibility.rst
```diff
@@ -19,18 +19,27 @@ libraries.
    :widths: 20 20 20
    :header-rows: 1
 
-   * - Ansys installation
+   * - DPF server version
      - ansys.dpf.core python module version
      - ansys.dpf.post python module version
-   * - Ansys 2022 R2
+   * - 6.1 (DPF Server 2023.2.pre1)
+     - 0.8.0 or later
```
JennaPaikowsky, line 26: Why the two bullets with the two different numbers? Wouldn't "0.4.0 or later" cover all versions after 0.4.0, including 0.8.0? If there is a reason for laying it out like this, please make it clearer.

JennaPaikowsky (commented):
Looks good, deleted my earlier comment as I forgot it was a table :)

PR #375, opened 2023-04-24
fix server update

cbellot000:


rafacanton (approved):

PProfizi:
Are we going to revert changes to "standalone_suffix" once the PR is ready?

cbellot000:
@PProfizi the post pipelines are broken with the current pre0 server because of a pip upgrade not supporting what we have in the setup.py, I will complete this PR and we can revert once the server is updated, what do you think?

PR #435, opened 2023-07-21
Add fluid result averaging, FluidMeshInfo, and Faces

PProfizi:
- Mark TODO: filter-out zones from `zone_ids` and `qualifiers` different from requested location for integrated results
- Mark TODO: Filter-out face zones from `zone_ids` or `qualifiers` when requesting results on cells 
- Mark TODO: Filter-out cell zones from `zone_ids` or `qualifiers` when requesting results on faces
- Implement averaging with `to_nodal_fc` (cells to nodes and faces to nodes)
- Implement averaging with `to_elemental_fc` (nodes to cells, nodes to faces)

Logic:
No averaging of integrated results anywhere else
No averaging between faces and cells
Averaging between nodes and cells
Averaging between faces and nodes


src/ansys/dpf/post/fluid_simulation.py
```diff
@@ -276,6 +285,10 @@ def _get_result(
                 "are mutually exclusive."
             )
 
+        if result_type == _result_type_enum.integrated:
+            # TODO accept only face zones
```
rafacanton, line 289: For now this is OK, but in the future we could also have integrated results on cells. For example, Heat generation on elements (in W), which would come from a heat volumetric density (W/m^3)

PProfizi: Hey @rafacanton, I modified the logic to use a `native_location`. This way when marked as `integrated=True` I only check that the location asked is the same as the native.

src/ansys/dpf/post/fluid_simulation.py
```diff
@@ -263,9 +260,8 @@ def _get_result(
             Overrides any other qualifier argument such as `phases`, `species` or `zone_ids`.
         named_selections:
             Named selection or list of named selections to get results for.
-        result_type:
-            A ``result_type_enum`` value to inform a test on requested zones is necessary.
-            An integrated result will for example only be available on faces.
+        integrated:
+            An integrated result cannot be requested on another location than its native_location.
```
rafacanton, line 264: Shouldn't we mention the actual definition of integrated result? Integrated results in Dpf are indeed able to be averaged. It's Fluids+integrated results the ones that aren't (to be compliant with FBU's directives)

src/ansys/dpf/post/fluid_simulation.py
```diff
@@ -1213,85 +1259,7 @@ def enthalpy_on_nodes(
             phases=phases,
             species=species,
             named_selections=named_selections,
-        )
-
-    def enthalpy_on_faces(
```
rafacanton, line 1181: Why removing this one¿

src/ansys/dpf/post/fluid_simulation.py
rafacanton: This makes sense to me, thanks!

src/ansys/dpf/post/fluid_simulation.py
PProfizi: @rafacanton this is an Elemental result from what I have seen. Which would prevent from averaging it onto faces?

src/ansys/dpf/post/fluid_simulation.py
PProfizi: @rafacanton could you give me a description of why FBU's directive is to not average integrated fluid results?

src/ansys/dpf/post/fluid_simulation.py
rafacanton: @PProfizi Because integrated results only make them in their "support". If you have a mass flow rate per face in a surface, you can directly sum all of them and have the overall mass flow rate that flows through the surface (say 20 kg/s). If you average the results to the nodes and then you numerically integrate it along the surface, you cannot guarantee that these 20 kg/s will be conserved. What you would need to do is average the intensive quantity (in this case density * v), average it to the nodes and then integrate it along the surface to get back mass flow rate.

src/ansys/dpf/post/fluid_simulation.py
```diff
@@ -218,14 +220,15 @@ def _get_result(
...
+            Native location of the result.
         location:
             Location to extract results at. Available locations are listed in
             class:`post.locations` and are: `post.locations.nodal`,
-            `post.locations.elemental`, and `post.locations.elemental_nodal`.
-            Using the default `post.locations.elemental_nodal` results in a value
-            for every node at each element. Similarly, using `post.locations.elemental`
-            gives results with one value for each element, while using `post.locations.nodal`
-            gives results with one value for each node.
+            `post.locations.elemental`, and `post.locations.faces`.
```
cbellot000, line 228: Could we add post.locations.cell to comply with fluid vocabulary? 
In the same topic, I'm wondering if we should have at least some docstrings mentioning "reconstruction" (the equivalent of "averaging" in fluid vocabulary) @rafacanton. It will help if people make a search in the doc

src/ansys/dpf/post/fluid_simulation.py
```diff
@@ -276,6 +281,45 @@ def _get_result(
                 "are mutually exclusive."
             )
 
+        # Raise for integrated result queried on anything else than its native location
+        if integrated:
+            if (native_location == "Faces" and location != locations.faces) or (
+                native_location == "Elements" and location != locations.elemental
```
cbellot000, line 287: why are we not using the locations strings in here?

PProfizi: @cbellot000 because they are not the native_location strings in the available results are not the same as what you can find in `locations`

src/ansys/dpf/post/fluid_simulation.py
rafacanton: @PProfizi Because that is what we were told to do. IMO we should also prevent it in solids, but that is just an opinion (and we have a specific algorithm in Dpf to average integrated results for FEM results, we don't do the same thing than we do for the others).

src/ansys/dpf/post/fluid_simulation.py
PProfizi: @cbellot000 @rafacanton the issue is there is no `gate.locations.cells`, yet `post.locations` comes from `core.locations` which comes from `gate.locations`.

tests/test_fluid_simulation.py
```diff
@@ -113,3 +108,315 @@ def test_fluid_simulation_phases(self, fluent_simulation):
...
+        result = cfx_simulation.density_on_nodes(
+            cell_ids=cfx_simulation.mesh.element_ids
+        )
+        assert result.index.mesh_index.location == post.locations.nodal
+        ref = """
+  results     RHO (kg*m^-3)           
+  set_ids                 1           
+    phase Water at 25 C (2) Copper (3)
+ node_ids                             
+     3149        9.9700e+02
```
rafacanton, line 219: The order is not the same as the previous one because the elements have been transposed to nodes, but if we are passing the whole set of elements in the mesh, the sizes should be. Should we check it?

tests/test_fluid_simulation.py
```diff
@@ -113,3 +108,315 @@ def test_fluid_simulation_phases(self, fluent_simulation):
...
+
+    def test_results_cfx_cross_locations_on_nodes(self, cfx_simulation):
+        result = cfx_simulation.density_on_nodes(node_ids=cfx_simulation.mesh.node_ids)
+        assert result.index.mesh_index.location == post.locations.nodal
+        ref = """
+  results     RHO (kg*m^-3)           
+  set_ids                 1           
+    phase Water at 25 C (2) Copper (3)
+ node_ids                             
+        1        9.9700e+02
```
rafacanton, line 201: Maybe density is not the best result because this seems to be an incompressible simulation (density is constant), so we will not be able to catch bugs regarding wrong scoping transposition)

tests/test_fluid_simulation.py
```diff
@@ -113,3 +108,315 @@ def test_fluid_simulation_phases(self, fluent_simulation):
...
+ face_ids                             
+"""  # noqa: W291, E501
+        assert str(result) == ref
+        with pytest.raises(
+            ValueError, match="Cannot plot a Dataframe with an empty mesh index."
+        ):
+            result.plot()
+
+    def test_results_cfx_cross_locations_on_cells(self, cfx_simulation):
+        result = cfx_simulation.density_on_cells(
```
rafacanton, line 274: Does it make sense to add an scenario of density_on_cells but with nodal scoping?

tests/test_fluid_simulation.py
```diff
@@ -113,3 +108,315 @@ def test_fluid_simulation_phases(self, fluent_simulation):
...
+      996    1.1041e+00
+      894    1.1035e+00
+      795    1.0982e+00
+      903    1.0985e+00
+      997    1.1097e+00
+      895    1.1091e+00
+      ...           ...
+"""  # noqa: W291, E501
+        assert str(result) == ref
+        result = fluent_simulation.density_on_nodes(
```
rafacanton, line 328: @PProfizi Let's say that we have an `ElementalAndFaces` result and cell 30, face 25 and node 15 connected. Can you confirm the following workflows?

1. Ask for `density_on_cells` with cell 30 -> this would return the value on the file -> the result will be elemental -> **OK**
2. Ask for `density_on_cells` with node 15 -> this would do a transposition nodes->cells + get the value from the file -> the result will be elemental -> **OK**
3. Ask for `density_on_cells` with face 25 -> this would do a transposition faces->cells + get the value from the file -> the result will be elemental -> **OK**.
4. Ask for `density_on_nodes` with cell 30 -> this would do a get the value from the file +averaging to the nodes -> the result will be nodal -> **OK**.
5. Ask for `density_on_nodes` with node 15 ->This one is tricky, as the result is ElementalAndFaces. This would in my mind do a transposition nodes->cells + get the value from the file + averaging to nodes -> the result will be nodal -> **OK**.
6. Ask for `density_on_nodes` with face 25 -> This would get the value from the file + average to nodes -> **OK, However, this is the bug and we forbid it**.
8. Ask for `density_on_faces` with cell 30 -> This would do a transposition to faces + get the value from the file ->the result will be in faces -> **OK**.
9. Ask for `density_on_faces` with node 15 -> this would do a transposition nodes->faces + get the value from the file -> **OK, However, this is the bug and we forbid it**.
10. Ask for `density_on_faces` with face 25 -> Get the value from the file -> **OK, However, this is the bug and we forbid it**.

tests/test_fluid_simulation.py
PProfizi: ok, I'll change for something else

tests/test_fluid_simulation.py
PProfizi: we do not plan to support nodal_scoping to cell_scoping for now, due to the question of inclusive or not, this may be added in a next version.
Sort of like for faces from cell_ids.

tests/test_fluid_simulation.py
PProfizi: @rafacanton 
1. Will add test
2. -> we do not currently allow querying cell result using node IDs
3. -> we do not currently allow querying cell result using face IDs
4. Will add test
5. Will check if this works
6. Forbidden as bugged
7. Will add test if not present
8. Not allowed for now
9. Forbidden as bugged

Basically I only allow querying results on "lower-dimension" entities with scopings, so cell_scoping for cell, face or nodal result, face_scoping for face or nodal result, node_scoping for nodal result

rafacanton: @PProfizi Thanks for sharing the logic! Now I have the following question: Suppose that we fix the server bug and the bugged scenarios are allowed. Case 4 will yield a value at node 15 that would average the surrounding cell values. However, Case 6 will yield a value at node 15 that would average the surrounding face values. They will differ in general. This is expected, but should be communicated in the example that demonstrates the feature

src/ansys/dpf/post/mesh_info.py
```diff
@@ -0,0 +1,192 @@
...
+
+"""
+from ansys.dpf import core as dpf
+
+
+class FluidMeshInfo:
+    """Holds the metadata relative to a fluid mesh.
+
+    This class describes the available mesh information.
+    Use this to gather information about a big mesh prior to loading it
```
MaxJPRey, line 14:
```suggestion
    Use this to gather information about a large mesh prior to loading it
```

src/ansys/dpf/post/faces.py
```diff
@@ -0,0 +1,158 @@
...
+        _str += "]"
+        return _str
+
+    def __str__(self) -> str:
+        """Returns a string representation of a _FaceList object."""
+        return self._short_list()
+
+    @abstractmethod
+    def __iter__(self):  # pragma: no cover
+        """Returns the object to iterate on."""
```
germa89, line 117: To add something to the function.

```suggestion
    def __iter__(self):  # pragma: no cover
        """Returns the object to iterate on."""
        raise NotImplemented
```

src/ansys/dpf/post/faces.py
```diff
@@ -0,0 +1,158 @@
...
+
+    def __len__(self) -> int:
+        """Returns the number of faces in the list."""
+        return self._face_list.n_faces
+
+    def __repr__(self) -> str:
+        """Returns a string representation of _FaceList object."""
+        return f"{self.__class__.__name__}({self}, __len__={len(self)})"
+
+    def _short_list(self) -> str:
```
germa89, line 99: Not sure that name is appropriate regarding what the function does...

cbellot000 (approved):

tests/test_fluid_simulation.py
PProfizi: @rafacanton see [here](https://github.com/ansys/pydpf-post/blob/f76a819e99bc43c3ed27e7a62774d2ec32676185/src/ansys/dpf/post/fluid_simulation.py#L336)

PR #445, opened 2023-07-28
Fix result query on zones

PProfizi:
Add the `MeshesProvider` operator to have meshes associated to the resulting fields when using a zone_scoping input

### DIFF
--- a/src/ansys/dpf/post/fluid_simulation.py
+++ b/src/ansys/dpf/post/fluid_simulation.py
@@ -388,12 +388,15 @@ class FluidSimulation(Simulation):
             location=location,
             force_elemental_nodal=False,
         )
+        query_regions_meshes = False
         lists = []
         lists_labels = []
         if qualifiers:
             labels = list(qualifiers.keys())
             lists_labels.extend(labels)
             lists.extend([qualifiers[key] for key in labels])
+            if "zone" in labels:
+                query_regions_meshes = qualifiers["zone"]
         else:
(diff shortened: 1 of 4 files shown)


src/ansys/dpf/post/fluid_simulation.py
```diff
@@ -423,6 +427,14 @@ def _get_result(
         # Its output is selected as future workflow output for now
         # print(result_op)
 
+        if query_regions_meshes:
+            # Results have been queried on regions,
+            # A MeshesProvider is required to give meshes as input of the source operator
+            meshes_provider_op = self._model.operator("meshes_provider")
+            meshes_provider_op.connect(25, query_regions_meshes)
+            wf.add_operator(meshes_provider_op)
+            result_op.connect(7, meshes_provider_op.outputs.meshes)
```
rafacanton, line 436:
```
if query_regions_meshes:
    # Results have been queried on regions,
    # A MeshesProvider is required to give meshes as input of the source operator
    meshes_provider_op = self._model.operator("meshes_provider")
    meshes_provider_op.connect(25, query_regions_meshes)
else:
    # Results have been queried on regions,
    # A MeshProvider is required to give meshes as input of the source operator
    meshes_provider_op = self._model.operator("mesh_provider")

wf.add_operator(meshes_provider_op)
result_op.connect(7, meshes_provider_op.outputs.meshes)

````

rafacanton (approved):

PR #449, opened 2023-08-01
Remove rogue prints from tests

PProfizi:


tests/test_fluid_simulation.py
```diff
@@ -110,12 +110,12 @@ def test_fluid_simulation_phases(self, fluent_simulation):
         assert isinstance(fluent_simulation.phases, PhasesDict)
 
     def test_fluid_simulation_result_unavailable(self, fluent_simulation):
-        print(fluent_simulation)
+        # print(fluent_simulation)
```
rafacanton, line 113: @PProfizi Are we then testing the string representation of the fluent_simulation in other context?

PR #462, opened 2023-08-11
Implement request of fluid results using faces

PProfizi:
Available with server 2024.1.pre1


tests/test_fluid_simulation.py
```diff
@@ -381,31 +389,38 @@ def test_results_fluent_cross_locations_on_faces(self, fluent_simulation):
...
+                )
+        else:
+            result = fluent_simulation.density_on_faces(
                 face_ids=fluent_simulation.mesh.face_ids
             )
+            print(result)
+            assert result.index.mesh_index.location == post.locations.faces
+            # assert (
+            #     len(result.index.mesh_index.values) == len(fluent_simulation.mesh.face_ids)
+            # )  # TODO: why does this fail? Is the result not defined everywhere?
```
rafacanton, line 409: @PProfizi Because we are not reading the results for interior faces. I have checked for the first cas file in the axial_comp model and I can confirm that:
- The overall mesh has 44242 faces.
- There are two interior zones that account for 38894 faces.
- The static_pressure result on faces has 5348 entities, which is exactly the same as 44242 - 38894.

PR #480, opened 2023-09-07
Expose streamlines

PProfizi:
Create the `ansys.dpf.post.tools` package, with the `ansys.dpf.post.tools.streamlines` module.
Add the `plot_streamlines` helper to provide a first way of plotting streamlines based on a DataFrame.
Add a first test in `test_streamlines.py`.

Initial mock-up:
image


src/ansys/dpf/post/tools/streamlines.py
```diff
@@ -0,0 +1,76 @@
...
+    dataframe: post.DataFrame,
+    sources: List[dict],
+    mesh: Union[post.Mesh, None] = None,
+    plot_mesh: bool = True,
+    mesh_opacity: float = 0.3,
+    plot_contour: bool = True,
+    contour_opacity: float = 0.3,
+    **kwargs,
+):
+    """Plot streamlines based on a velocity field DataFrame.
```
rafacanton, line 25:
```suggestion
    """Plot streamlines based on a vector field DataFrame.
```

src/ansys/dpf/post/tools/streamlines.py
```diff
@@ -0,0 +1,76 @@
...
+
+    Parameters
+    ----------
+    dataframe:
+        A `post.DataFrame` object containing a velocity field.
```
rafacanton, line 30:
```suggestion
        A `post.DataFrame` object containing a vector field.
```

src/ansys/dpf/post/tools/streamlines.py
```diff
@@ -0,0 +1,76 @@
...
+    sources:
+        A list of dictionaries defining spherical point sources for the streamlines.
+        Expected keywords are "center", "radius", and "line_thickness".
+    mesh:
+        A `post.Mesh` object to use to compute the streamlines instead of the mesh associated with
+        the velocity field in the DataFrame.
```
rafacanton, line 36:
```suggestion
        the vector field in the DataFrame.
```

src/ansys/dpf/post/tools/streamlines.py
```diff
@@ -0,0 +1,76 @@
...
+    plot_contour:
+        Whether to plot the field contour along with the streamlines.
```
rafacanton, line 38:
```suggestion
        Whether to plot the field's norm contour along with the streamlines.
```

rafacanton (commented):
@PProfizi Just made a couple of suggestions to avoid referring to velocity. This is way more general and applies to all vector fields. For example, when we support EMag data, the same method could be used to plot the magnetic field lines

src/ansys/dpf/post/tools/streamlines.py
```diff
@@ -0,0 +1,76 @@
...
+    mesh_opacity:
+        Opacity to use for the mesh in case "plot_contour=False" and "plot_mesh=True".
+    **kwargs:
+
+    """
+    # Select data to work with
+    if mesh:
+        meshed_region = mesh._meshed_region
+    else:
+        meshed_region = dataframe._fc[0].meshed_region
```
cbellot000, line 52: we will eventually need a public method in the dataframe to get the mesh, no?

src/ansys/dpf/post/tools/streamlines.py
```diff
@@ -0,0 +1,76 @@
...
+    field = dataframe._fc[0]
```
cbellot000, line 53: why the first field? what if you have a single time, but different label (like "elshape") in your data frame?

src/ansys/dpf/post/tools/streamlines.py
PProfizi: @cbellot000 sure, but again it will need a label logic to select which field's meshed_region is of interest, or else I guess return a `MeshesContainer` as a `Meshes`

rafacanton (approved):

PProfizi:
@rafacanton @cbellot000 @anslpa 
Note here to keep the information:

Cannot actually create a `ansys.dpf.post.tools` sub-package with `ansys.dpf.post.tools.streamlines` module (`ansys/dpf/post/tools/streamlines.py`) as we already have a `ansys.dpf.post.tools` module (`ansys/dpf/post/tools.py`).
Three solutions:
- rename the new sub-package from `tools` to `helpers`. This is the current solution implemented, so now the new method is in `ansys.dpf.post.helpers.streamlines`. This is also what was done in `ansys-dpf-core` which makes both libraries coherent.
- Move the `streamlines.py` module directly in `ansys/dpf/post`
- Move the streamlines helpers in the `tools.py` module, which would mean no streamlines-specific module and a direct call to `from ansys.dpf.post.tools import plot_streamlines`.
- Remove the `tools.py` module and move its methods in a new module in the `tools` sub-package. This would mean makign sure that the old imports would still work. Not sure we can cover all edge-cases.

I think the first solution is the best.

What do you think?

rafacanton:
> @rafacanton @cbellot000 @anslpa
> 
> Just realized we cannot actually create a `ansys.dpf.post.tools` sub-package with `ansys.dpf.post.tools.streamlines` module (`ansys/dpf/post/tools/streamlines.py`) as we already have a `ansys.dpf.post.tools` module (`ansys/dpf/post/tools.py`). Four solutions:
> 
> * rename the new sub-package from `tools` to `helpers`. This is the current solution implemented, so now the new method is in `ansys.dpf.post.helpers.streamlines`. This is also what was done in `ansys-dpf-core` which makes both libraries coherent.
> * Move the `streamlines.py` module directly in `ansys/dpf/post`
> * Move the streamlines helpers in the `tools.py` module, which would mean no streamlines-specific module and a direct call to `from ansys.dpf.post.tools import plot_streamlines`.
> * Remove the `tools.py` module and move its methods in a new module in the `tools` sub-package. This would mean makign sure that the old imports would still work. Not sure we can cover all edge-cases.
> 
> I think the first solution is the best.
> 
> What do you think?

@PProfizi I agree, I think it is right to have both libraries with a similar structure. Thanks!

PProfizi:
> > @rafacanton @cbellot000 @anslpa
> > Just realized we cannot actually create a `ansys.dpf.post.tools` sub-package with `ansys.dpf.post.tools.streamlines` module (`ansys/dpf/post/tools/streamlines.py`) as we already have a `ansys.dpf.post.tools` module (`ansys/dpf/post/tools.py`). Four solutions:
> > 
> > * rename the new sub-package from `tools` to `helpers`. This is the current solution implemented, so now the new method is in `ansys.dpf.post.helpers.streamlines`. This is also what was done in `ansys-dpf-core` which makes both libraries coherent.
> > * Move the `streamlines.py` module directly in `ansys/dpf/post`
> > * Move the streamlines helpers in the `tools.py` module, which would mean no streamlines-specific module and a direct call to `from ansys.dpf.post.tools import plot_streamlines`.
> > * Remove the `tools.py` module and move its methods in a new module in the `tools` sub-package. This would mean makign sure that the old imports would still work. Not sure we can cover all edge-cases.
> > 
> > I think the first solution is the best.
> > What do you think?
> 
> @PProfizi I agree, I think it is right to have both libraries with a similar structure. Thanks!

What I ended-up doing:
- add a new `ansys.dpf.post.helpers` sub-package
- add a new `ansys.dpf.post.helpers.streamlines` module
- add a new `ansys.dpf.post.helpers.selections` module
- moved the `ansys.dpf.post.tools.select` method to the `ansys.dpf.post.helpers.selections` module
- made an import to it in `ansys.dpf.post.tools` to ensure retro-compatibility
- updated the dosctring in `ansys.dpf.post.tools` to mention the move to `ansys.dpf.post.helpers`

src/ansys/dpf/post/helpers/__init__.py
```diff
@@ -0,0 +1,6 @@
+"""Tools package.
+
+Tools
```
cbellot000, line 3: Helpers?

cbellot000: Is there a plan for follow up PR adding an example for streamlines?

PProfizi: > The only problem I see is naming. For me "helpers" means wrapping already existing capabilities in another way to make them easier to be used

IMO this is what this method is? The end-goal is to have a much more thorough streamlines manipulation capability, this method being a quick way to plot the streamlines given a Dataframe.



> Is there a plan for follow up PR adding an example for streamlines?

Yes of course, I initially wanted this merged before last Friday's deadline, but if the deadline is not an issue anymore, I can add and example within this PR.

PProfizi: > The only problem I see is naming. For me "helpers" means wrapping already existing capabilities in another way to make them easier to be used

@cbellot000 What I can do though is rename `helpers` as `tools`, but that will require deleting the `tools.py` module, which I think can be done without retro-compatibility issues if I import the `select` method from within the `__init__.py` file of the new `tools` sub-package, yet I may have missed a case where it breaks.

cbellot000: @PProfizi no, it's okay, your justification makes sense

cbellot000 (approved):

PR #516, opened 2023-10-16
Remove Vale Google.Headings warnings due to false positives

PProfizi:
to prevent the latest issues with errors wrongly raised.


PProfizi:
HI @Revathyvenugopal162 ,
Here is a PR where I try to make the Vale false warnings go away, and they still appear randomly (see multiple tries of the CI, just rerunning the `Check doc style` job).
It seems this is a current problem for everyone using Vale.
Pinning to `2.29.3` did not solve the issue.

rafacanton (approved):

PR #519, opened 2023-10-19
Clean init file

PProfizi:
Remove call to `core.settings.set_dynamic_available_results_capability(False)`

@cbellot000 do you remember why this was put there? It dates back to before the available GH/git history.

@FedericoNegri reported it being there and not necessarily being practical for the user in some situations.

### DIFF
--- a/src/ansys/dpf/post/__init__.py
+++ b/src/ansys/dpf/post/__init__.py
@@ -55,7 +55,3 @@ from ansys.dpf.post.transient_mechanical_simulation import (  # noqa: F401
 
 # this must be after some ansys.dpf.post import
 __version__ = importlib_metadata.version("ansys-dpf-post")
-
-if hasattr(core, "settings") and hasattr(core.settings, "set_default_pyvista_config"):
-    core.settings.set_default_pyvista_config()
-# dpf.core.start_local_server()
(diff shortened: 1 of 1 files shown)


cbellot000 (approved):

cbellot000:
> Remove call to `core.settings.set_dynamic_available_results_capability(False)`
> 
> @cbellot000 do you remember why this was put there? It dates back to before the available GH/git history.
> 
It's an API from pydpf-core? It would have been with the legacy API. But I'm intesrested to see where it's used in core?

PProfizi:
> > Remove call to `core.settings.set_dynamic_available_results_capability(False)`
> > @cbellot000 do you remember why this was put there? It dates back to before the available GH/git history.
> 
> It's an API from pydpf-core? It would have been with the legacy API. But I'm intesrested to see where it's used in core?

It is in `ansys.dpf.core.settings`, and it switches the variable `misc.DYNAMIC_RESULTS` which is used in `Model.results` [here](https://github.com/ansys/pydpf-core/blob/6d945d52cbceb67f52ad6f2b4deeb1faa6a16205/src/ansys/dpf/core/model.py#L161) to force having only `CommonResults`.

It was dedicated to the legacy PyDPF-Post, but since it is in the init, it is the case also for the new Post API.

PR #543, opened 2023-11-15
Expose workflow to extract results from simulation

FedericoNegri:


cbellot000 (approved):

src/ansys/dpf/post/harmonic_mechanical_simulation.py
```diff
@@ -334,6 +258,131 @@ def _get_result(
             comp = None
             base_name += "_N"
 
+        return wf, out, selection
```
PProfizi, line 261: @FedericoNegri why do you need to set the "out" pin of the workflow later and not here?

FedericoNegri: @PProfizi In a different context, I'm using the `_get_result_workflow` function to get the workflow and chain additional operators, for instance:

```python

wf, out = _get_result_workflow(...)

# add min/max filters
min_max_op = simulation._model.operator(name="min_max_fc")
min_max_op.connect(0, out)
wf.add_operator(operator=min_max_op)

# Set the workflow output
wf.set_output_name("min", min_max_op.outputs.field_min)
wf.set_output_name("max", min_max_op.outputs.field_max)
```

Do you suggest doing differently?

PProfizi: @FedericoNegri @cbellot000 I've had the same sort of issues while trying to make dynamic workflows, I think what we need is to add getter to `Workflow`, such as `Workflow.get_output_by_name` which returns the input/ouput as a proper `Operator.Input` or `Operator.Output` the same way we can `set_output_name` or `set_input_name`, otherwise we keep having to pass a variable to not loose it. The alternative is to use `Workflow.connect_with` but it's not always very user-friendly.

FedericoNegri: @PProfizi thanks for the feedback. I can try out the `connect_with` route, but yeah ideally the first approach would be more convenient. Anyways, I've reworked the `_get_result_workflow` so that now it only returns the workflow. Please have a look.

PR #615, opened 2024-03-29
Add FluidSimulation.zone_mesh() to get the mesh of a single zone

PProfizi:


src/ansys/dpf/post/fluid_simulation.py
```diff
@@ -177,6 +178,15 @@ def mesh_info(self) -> FluidMeshInfo:
             self._mesh_info = FluidMeshInfo(self._model.metadata.mesh_info)
         return self._mesh_info
 
+    def zone_mesh(self, zone_id: int) -> Mesh:
```
cbellot000, line 181: @PProfizi can we accept a zone name as well please? This would make it more consistent with the requests on phases

src/ansys/dpf/post/fluid_simulation.py
```diff
@@ -177,6 +178,28 @@ def mesh_info(self) -> FluidMeshInfo:
             self._mesh_info = FluidMeshInfo(self._model.metadata.mesh_info)
         return self._mesh_info
 
+    def zone_mesh(self, zone: Union[int, str]) -> Mesh:
```
cbellot000, line 181: @PProfizi don't you think we should create a mesh api container here ? for extensibility of our APIs I think it can be useful. We could need to add other APIs like: zone_meshes(zones), merged_mesh_for_zones()...

PProfizi: I've marked this as a global feature request for mesh manipulation in PyDPF-Post.
This one though was a specific request so I'll merge it on its own for now.
To me this method will still be there, just improved (for example allow to select several zones at once and get a single Mesh.

PR #673, opened 2024-07-30
Fix combination of skin extraction with scoping

### LINKED ISSUE #423: Issues combining mesh skin with scoping
### Before submitting the issue

- [X] I have checked for [Compatibility](https://dpf.docs.pyansys.com/version/stable/getting_started/index.html#compatibility) issues
- [X] I have searched among the existing issues
- [X] I am using a Python virtual environment

### Description of the bug

I'm playing with scoping in combination with skin mesh, and see something unexpected.
In this script
```python
from ansys.dpf import post

rst_path = r"D:\beam_2_mat.rst"

simulation = post.StaticMechanicalSimulation(rst_path)
simulation._model.metadata.meshed_region.plot()

# get an elemental named selection
scoping = simulation._model.metadata.meshed_region.named_selection(
    'X_MIN_SOLID_ELE'
)

seqv_ns_full = simulation.stress_eqv_von_mises_elemental(element_ids=scoping.ids)
seqv_ns_full.plot()

seqv_ns_skin = simulation.stress_eqv_von_mises_elemental(skin=scoping.ids)
seqv_ns_skin._fc[0].meshed_region.plot()

seqv_ns_skin.plot()
```

I get this (correct) plot for the field on the solid mesh
MicrosoftTeams-image

but this one on the skin
MicrosoftTeams-image (1)

The same rst file also contains other named selections that can be used for comparison. For instance `MG_BODY_ELE` and `MG_BODY_NODES` are named selections covering the same region with elemental and nodal location, respectively. The same applies to `ALU_BODY_ELE` and `ALU_BODY_NODES`.

Would be good to validate these kind of comparisons:

```python
# get an elemental named selection
scoping = simulation._model.metadata.meshed_region.named_selection(
    'MG_BODY_ELE'
)

seqv_ns_full = simulation.stress_eqv_von_mises_elemental(element_ids=scoping.ids)
seqv_ns_full.plot()

seqv_ns_skin = simulation.stress_eqv_von_mises_elemental(skin=scoping.ids)
seqv_ns_skin._fc[0].meshed_region.plot()
seqv_ns_skin.plot()

# get a nodal named selection
scoping = simulation._model.metadata.meshed_region.named_selection(
    'MG_BODY_NODES'
)

seqv_ns_full = simulation.stress_eqv_von_mises_nodal(node_ids=scoping.ids)
seqv_ns_full.plot()

seqv_ns_skin = simulation.stress_eqv_von_mises_nodal(skin=True, node_ids=scoping.ids)
seqv_ns_skin._fc[0].meshed_region.plot()
seqv_ns_skin.plot()
```

The rst file: [beam_2_mat.zip](https://github.com/ansys/pydpf-post/files/11818539/beam_2_mat.zip)

### Steps To Reproduce

-

### Which Operating System are you using?

Windows

### Which DPF/Ansys version are you using?

Ansys 2023 R2

### Which Python version are you using?

3.10

### Installed packages

-

janvonrickenbach:
fixes #423

This PR fixes the following problem: The skin mesh operator stores the mapping of skin elements to solid elements as element indices (instead of element ids). The solid_to_skin_fc operator by default relies on these indices for the mapping of the solid data onto the skin mesh. This means that solid_to_skin_fc operator only works correctly if it is passed the same mesh that was the input for the skin operator. There is a second input to the solid_to_skin_fc operator which allows to pass the mesh on which the skin extraction happened (as opposed to the support of the result field). If I pass the correct mesh there, the data is correct for the model attached to the bug.

Todo:

- [x] Discuss why equivalent strains are computed before the averaging
- [x] Verify cyclic results

Some expectations have changed, but I believe the old results were incorrect. Here is an example results before the changes:

test_simulation py_TestHarmonicMechanicalSimulation_test_stress_skin (call)_

And this is the result with the fixes:
test_simulation py_TestHarmonicMechanicalSimulation_test_stress_skin (call)_skin_eqv

Here is the updated workflow: 
workflow_0x0000019BC2ACAF20

and this was the old workflow:
workflow_0x000002E2E3A3A0B0

The two workflows are essentially the same, I just added the "skin_input_mesh" output. Also some forwarding operators were added to simplify the code.

I added extensive tests which uncovered the following issues (see more details in the tests). I would not address these issues in this PR as their resolution needs to be further discussed.

1) The principal strains are computed incorrectly in the skin workflow, because the off-diagonal elements are not multiplied by 0.5. This depends on a "strain" field property which is not propagated correctly in the skin workflow. I guess this can be fixed by correctly forwarding the "strain" property in the skin operators. Note that this cannot be fixed in python since the field properties are not exposed. Fixed in backend (solid to skin operator now propagates the header).

2) Skin extraction for elemental nodal data does currently not work. The workflow just returns the elemental nodal data of the solid in this case.  Created issue #685

3) When comparing nodal results on the scoped skin to the corresponding data of the complete mesh, there are discrepancies at the boundaries. In the skin workflow the elemental nodal data is scoped and then averaged. This means that at the boundaries of the scope, only the  elemental nodal data "inside" the scope is taken into account for the nodal average.  => this is considered ok tests are adapted to reproduce this behavior

4) Another discrepancy comes from the fact that the averaging happens on the skin mesh. This means that in corners the adjacent elements get different weights depending on the configuration. Consider the example below.
The average of the circled node computed on the skin is the elemental nodal value of skin elements (black numbers 1,2,3,4,5) for this node. So the elements 1 and 3 (yellow numbers) are considered twice in the averaging, because they have two faces that touch the node. When averaging on the solids, the node would just be the average of the elemental nodal value of the three solid elements. => this is considered ok, tests are adapted to reproduce this behavior

5) Skin extraction does not work out of the box if the result file contains different element types apart from solid elements. In this case the skin input needs to be rescoped to the solid elements. Created bug: https://github.com/ansys/pydpf-post/issues/686

Both points 3) and 4) could be resolved by first averaging the data and then extracting the skin / element scope. However this would imply a performance penalty. We could also do the averaging on the external layer only. The values at the boundaries can only be correct if an additional layer of elements "around" the element scope is included in the averaging.

image

### DIFF
--- a/src/ansys/dpf/post/harmonic_mechanical_simulation.py
+++ b/src/ansys/dpf/post/harmonic_mechanical_simulation.py
@@ -125,6 +125,8 @@ class HarmonicMechanicalSimulation(MechanicalSimulation):
             equivalent_op = self._model.operator(name="eqv_fc")
             wf.add_operator(operator=equivalent_op)
             # If a strain result, change the location now
+            # TBD: Why do we put the the equivalent operator
+            # before the averaging operator for strain results?
             if (
                 average_op is not None
                 and category == ResultCategory.equivalent
--- a/src/ansys/dpf/post/selection.py
+++ b/src/ansys/dpf/post/selection.py
@@ -406,11 +406,12 @@ class SpatialSelection:
             on the expanded mesh.
         """
         def connect_any(input, input_value):
+            # Workaround to connect any inputs: see
+            # https://github.com/ansys/pydpf-core/issues/1670
             input._operator().connect(input._pin, input_value)
 
         skin_operator = operators.mesh.skin(server=self._server)
         self._selection.add_operator(skin_operator)
-
 
         initial_mesh_fwd_op = operators.utility.forward(server=self._server)
         self._selection.set_input_name(_WfNames.initial_mesh, initial_mesh_fwd_op.inputs.any)
--- a/src/ansys/dpf/post/selection.py
+++ b/src/ansys/dpf/post/selection.py
@@ -414,9 +414,11 @@ class SpatialSelection:
 
         initial_mesh_fwd_op = operators.utility.forward()
         self._selection.set_input_name(_WfNames.initial_mesh, initial_mesh_fwd_op.inputs.any)
+        self._selection.add_operator(initial_mesh_fwd_op)
 
         skin_operator_input_mesh_fwd_op = operators.utility.forward()
         connect_any(skin_operator_input_mesh_fwd_op.inputs.any, initial_mesh_fwd_op)
+        self._selection.add_operator(skin_operator_input_mesh_fwd_op)
 
         if _is_model_cyclic(is_model_cyclic):
             mesh_provider_cyc = operators.mesh.mesh_provider()
@@ -498,6 +500,8 @@ class SpatialSelection:
         self._selection.set_output_name(
             _WfNames.skin_input_mesh, skin_operator_input_mesh_fwd_op.outputs.any
         )
+
+        self._selection.view()
 
     def select_with_scoping(self, scoping: Scoping):
         """Directly sets the scoping as the spatial selection.
--- a/src/ansys/dpf/post/selection.py
+++ b/src/ansys/dpf/post/selection.py
@@ -33,6 +33,8 @@ from numpy import ndarray
 class _WfNames:
     data_sources = "data_sources"
     scoping = "scoping"
+    skin_scoping = "skin_scoping"
+    skin_input_mesh = "skin_input_mesh"
     final_scoping = "final_scoping"
     scoping_a = "scoping_a"
     scoping_b = "scoping_b"
@@ -408,6 +410,8 @@ class SpatialSelection:
         self._selection.add_operator(op)
         mesh_input = op.inputs.mesh
 
+        solid_mesh_output = None
+
         if _is_model_cyclic(is_model_cyclic):
             mesh_provider_cyc = operators.mesh.mesh_provider()
             self._selection.add_operator(mesh_provider_cyc)
@@ -437,8 +441,11 @@ class SpatialSelection:
                 )
                 self._selection.add_operator(mesh_by_scop_op)
                 op.inputs.mesh.connect(mesh_by_scop_op)
+                solid_mesh_output = mesh_by_scop_op.outputs.mesh
             else:
                 op.inputs.mesh.connect(mesh_provider_cyc)
+                solid_mesh_output = mesh_provider_cyc.outputs.mesh
+
             self._selection.set_input_name(
                 _WfNames.initial_mesh, mesh_provider_cyc, 100
             )  # hack
@@ -455,6 +462,16 @@ class SpatialSelection:
             self._selection.add_operator(mesh_by_scop_op)
             mesh_input = mesh_by_scop_op.inputs.mesh
             op.inputs.mesh.connect(mesh_by_scop_op)
+            solid_mesh_output = mesh_by_scop_op.outputs.mesh
+            fwd_scoping_op = operators.utility.forward(elements, server=self._server)
+            self._selection.set_output_name(
+                _WfNames.scoping, fwd_scoping_op.outputs.any
+            )
+
+        else:
+            input_mesh_fwd_op = operators.utility.forward()
+            self._selection.set_input_name(_WfNames.initial_mesh, input_mesh_fwd_op.inputs.any)
+            solid_mesh_output = input_mesh_fwd_op.outputs.any
 
         if mesh_input is not None:
             self._selection.set_input_name(_WfNames.initial_mesh, mesh_input)
@@ -463,7 +480,7 @@ class SpatialSelection:
         self._selection.set_output_name(_WfNames.skin, op.outputs.mesh)
         if location == locations.nodal and result_native_location == locations.nodal:
             self._selection.set_output_name(
-                _WfNames.scoping, op.outputs.nodes_mesh_scoping
+                _WfNames.skin_scoping, op.outputs.nodes_mesh_scoping
             )
 
         elif not _is_model_cyclic(is_model_cyclic) and (
@@ -478,8 +495,17 @@ class SpatialSelection:
                 _WfNames.initial_mesh, transpose_op.inputs.meshed_region
             )
             self._selection.set_output_name(
-                _WfNames.scoping, transpose_op.outputs.mesh_scoping_as_scoping
+                _WfNames.skin_scoping, transpose_op.outputs.mesh_scoping_as_scoping
             )
+
+        # Provide the input mesh from which a skin was generated
+        # This is useful because the skin_mesh contains the mapping of
+        # skin elements to the original mesh element indices, which is used
+        # by the solid_to_skin_fc operator. The skin_input_mesh can be passed
+        # to the solid_to_skin_fc operator to ensure that the mapping is correct.
+        self._selection.set_output_name(
+            _WfNames.skin_input_mesh, solid_mesh_output
+        )
 
     def select_with_scoping(self, scoping: Scoping):
         """Directly sets the scoping as the spatial selection.
--- a/src/ansys/dpf/post/selection.py
+++ b/src/ansys/dpf/post/selection.py
@@ -501,8 +501,6 @@ class SpatialSelection:
             _WfNames.skin_input_mesh, skin_operator_input_mesh_fwd_op.outputs.any
         )
 
-        self._selection.view()
-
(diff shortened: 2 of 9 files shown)


src/ansys/dpf/post/selection.py
```diff
@@ -404,9 +405,20 @@ def select_skin(
...
-        op = operators.mesh.skin(server=self._server)
-        self._selection.add_operator(op)
-        mesh_input = op.inputs.mesh
+        def connect_any(input, input_value):
+            input._operator().connect(input._pin, input_value)
+
+        skin_operator = operators.mesh.skin(server=self._server)
+        self._selection.add_operator(skin_operator)
+
+
```
janvonrickenbach, line 414: Add forwarding operators so we can simplify the routing.

tests/test_simulation.py
```diff
@@ -2007,21 +2007,43 @@ def test_disp_skin(self, frame_modal_simulation: post.ModalMechanicalSimulation)
...
-        if SERVERS_VERSION_GREATER_THAN_OR_EQUAL_TO_7_1:
-            assert len(result.index.mesh_index) == 36
+            result_full = frame_modal_simulation.stress_elemental(all_sets=True, skin=True)
+            assert len(result_full.index.mesh_index) == 11146
+            assert len(result_full.columns.set_ids) == 6
+
+        element_ids = list(range(1, 100))
+        result_skin_scoped = frame_modal_simulation.stress_elemental(all_sets=True, skin=element_ids)
+        result_solid_scoped = frame_modal_simulation.stress(all_sets=True, element_ids=element_ids)
+
```
janvonrickenbach, line 2022: Add skin consistency checks for raw results.

tests/test_simulation.py
```diff
@@ -2071,6 +2093,7 @@ def test_stress_skin2(self, frame_modal_simulation: post.ModalMechanicalSimulati
 
     def test_strain_skin(self, frame_modal_simulation: post.ModalMechanicalSimulation):
         if frame_modal_simulation._model._server.meet_version("7.1"):
+            # todo I think this should be strain instead of stress
             result = frame_modal_simulation.stress_principal_elemental(
```
janvonrickenbach, line 2097: TBD: There  are some tests labeled test_strain_skin, but they actually test stress results. I think this should be changed.

src/ansys/dpf/post/harmonic_mechanical_simulation.py
```diff
@@ -125,6 +125,8 @@ def _get_result_workflow(
             equivalent_op = self._model.operator(name="eqv_fc")
             wf.add_operator(operator=equivalent_op)
             # If a strain result, change the location now
+            # TBD: Why do we put the the equivalent operator
+            # before the averaging operator for strain results?
```
PProfizi, line 129: Hey @janvonrickenbach,
If I remember correctly, it should be because this is what is done in the Mechanical workflow.

PProfizi:
Hi @janvonrickenbach, this PR looks great to me! We do indeed need to create separate issues for what you uncovered. Thanks!

PProfizi (approved):

src/ansys/dpf/post/harmonic_mechanical_simulation.py
janvonrickenbach: Thanks, this is also what Camille mentioned. I will add a note.

FedericoNegri (commented):
Thanks for addressing this, the detailed explanation and follow ups on those other issues you found. I've installed your branch in our app and could verify that the original issue is fixed, it works as intended now.

While testing it, I did notice that it still takes an unexpectedly large amount of time to extract results on skin scoped to a named selection. On the same model I shared in the bug description, it takes <0.1s on the full mesh vs 4 seconds on the skin! 

I looked into the dpf debug log and it seems like most of the time is spent in the `by_scoping` operator

image
which I believe we're calling in SpatialSelection.select_skin (the operator is called `from_scoping` in Python). (I already reported in a different context findings of the `by_scopings` operator being too slow.) 

I know this is unrelated to this PR, but I'd like to get confirmation of the timings I get and double check that using that operator is the right approach. We can then follow up accordingly.

janvonrickenbach:
> Thanks for addressing this, the detailed explanation and follow ups on those other issues you found. I've installed your branch in our app and could verify that the original issue is fixed, it works as intended now.
> 
> While testing it, I did notice that it still takes an unexpectedly large amount of time to extract results on skin scoped to a named selection. On the same model I shared in the bug description, it takes <0.1s on the full mesh vs 4 seconds on the skin!
> 
> I looked into the dpf debug log and it seems like most of the time is spent in the `by_scoping` operator
> 
> image which I believe we're calling in SpatialSelection.select_skin (the operator is called `from_scoping` in Python). (I already reported in a different context findings of the `by_scopings` operator being too slow.)
> 
> I know this is unrelated to this PR, but I'd like to get confirmation of the timings I get and double check that using that operator is the right approach. We can then follow up accordingly.

I can confirm that the by_scoping operator by far the bottleneck in the "skin with element subset" workflow. For me the skin workflow with all elements selected (skin=True) is very fast (<1 ms). However, getting the full result is also relatively slow (see below). I'm not sure why the "S" operator is slow while the "mapdl::rst:S" operator is does not take any time @PProfizi @cbellot000 Could you comment on these numbers and possible alternative solutions?

```
TIME REPORT
##########################################################################################################################

                                                 method name      number of call     total time (ms)    avg time by call
------------------------------------------------------------------------------------------------------------------------
                                                       :Id-1                   1                   0                   0
                                                      INPUTS                  10                   0                   0
                                                     OUTPUTS                  10                   0                   0
                                     ResultInfoProvider:Id19                   1                   0                   0
                                     ResultInfoProvider:Id21                   1                   0                   0
                                                      S:Id16                   1                3228                3228
                                           WORKFLOW RUN(out)                   1                3229                3229
                                      entity_average_fc:Id25                   1                   1                   1
                                                forward:Id10                   1                   0                   0
                         mapdl::rst::ResultInfoProvider:Id20                   1                   0                   0
                         mapdl::rst::ResultInfoProvider:Id22                   1                   0                   0
                                          mapdl::rst::S:Id23                   1                   1                   1
                                        to_elemental_fc:Id24                   1                   3                   3
##########################################################################################################################
```

cbellot000:
> ##################################################

On the top of my mind, I don't see workarounds. We need to investigate the performance issue on mesh by scoping(s), FYI @rafacanton

FedericoNegri (approved):

janvonrickenbach:
> > Thanks for addressing this, the detailed explanation and follow ups on those other issues you found. I've installed your branch in our app and could verify that the original issue is fixed, it works as intended now.
> > While testing it, I did notice that it still takes an unexpectedly large amount of time to extract results on skin scoped to a named selection. On the same model I shared in the bug description, it takes <0.1s on the full mesh vs 4 seconds on the skin!
> > I looked into the dpf debug log and it seems like most of the time is spent in the `by_scoping` operator
> > image which I believe we're calling in SpatialSelection.select_skin (the operator is called `from_scoping` in Python). (I already reported in a different context findings of the `by_scopings` operator being too slow.)
> > I know this is unrelated to this PR, but I'd like to get confirmation of the timings I get and double check that using that operator is the right approach. We can then follow up accordingly.
> 
> I can confirm that the by_scoping operator by far the bottleneck in the "skin with element subset" workflow. For me the skin workflow with all elements selected (skin=True) is very fast (<1 ms). However, getting the full result is also relatively slow (see below). I'm not sure why the "S" operator is slow while the "mapdl::rst:S" operator is does not take any time @PProfizi @cbellot000 Could you comment on these numbers and possible alternative solutions?
> 
> ```
> TIME REPORT
> ##########################################################################################################################
> 
>                                                  method name      number of call     total time (ms)    avg time by call
> ------------------------------------------------------------------------------------------------------------------------
>                                                        :Id-1                   1                   0                   0
>                                                       INPUTS                  10                   0                   0
>                                                      OUTPUTS                  10                   0                   0
>                                      ResultInfoProvider:Id19                   1                   0                   0
>                                      ResultInfoProvider:Id21                   1                   0                   0
>                                                       S:Id16                   1                3228                3228
>                                            WORKFLOW RUN(out)                   1                3229                3229
>                                       entity_average_fc:Id25                   1                   1                   1
>                                                 forward:Id10                   1                   0                   0
>                          mapdl::rst::ResultInfoProvider:Id20                   1                   0                   0
>                          mapdl::rst::ResultInfoProvider:Id22                   1                   0                   0
>                                           mapdl::rst::S:Id23                   1                   1                   1
>                                         to_elemental_fc:Id24                   1                   3                   3
> ##########################################################################################################################
> ```

For future reference: solid workflow was slow because of the license checkout.

PR #678, opened 2024-08-05
Fix concurrent artifact-upload error 

PProfizi:
by requiring wheel upload for only one configuration (since os-agnostic)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -81,6 +81,7 @@ jobs:
           MODULE: ${{env.MODULE}}
           dpf-standalone-TOKEN: ${{secrets.DPF_PIPELINE}}
           install_extras: plotting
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -81,7 +81,7 @@ jobs:
           MODULE: ${{env.MODULE}}
           dpf-standalone-TOKEN: ${{secrets.DPF_PIPELINE}}
           install_extras: plotting
-          wheel: true
+          wheel: ${{ (matrix.python-version == env.MAIN_PYTHON_VERSION) && (matrix.os == 'windows-latest') }}
```
BClappe, line 84: Why do we only want to ceate a wheel for windows?
Don't we want one for Linux?

BClappe (approved):
May you write somewhere (in the pipeline would be good) why you do this and that the package is os-agnostic and this is why only windows is selected?

PR #684, opened 2024-08-20
Refactor and unify result workflows

janvonrickenbach:
This PR is pure refactoring of the result workflows in dpf post. 

The goal is to centralize building the workflow for the different simulation types. Before this PR there was a lot of code duplication in the different simulation types. The old code was quite hard to follow, because the creation of the sub-workflows and their connection was mixed. To simplify the code I took the following approach:

* Convert all the sub-steps (averaging, principal, equivalent etc) into workflows. This simplifies the connection of the different sub-workflows. Previously a mix of operators, list of operators and (sub)workflows was used.
* The creation of the overall workflow is split into two steps: 
    - Create all the sub-workflows and put them into a ResultWorkflows object. sub-workflows that are not needed are None,
    - Connect the sub-workflows to get a single big workflow
* Everything is moved to free functions so it is easy to reuse different parts of the workflow. 
* Moved component related logic from Simulation class to free functions.


src/ansys/dpf/post/fluid_simulation.py
```diff
@@ -223,16 +236,43 @@ def _get_result_workflow(
         qualifiers: Union[dict, None] = None,
     ) -> (dpf.Workflow, Union[str, list[str], None], str):
         """Generate (without evaluating) the Workflow to extract results."""
```
janvonrickenbach, line 238: I took some code out of the old _build_result_workflow. The assumption here is that cyclic expansion is not supported for fluid simulations. I cannot use the standard initial_result_workflow because the operator needs to be exposed and is used on line 311.

src/ansys/dpf/post/result_workflows/_component_helper.py
```diff
@@ -0,0 +1,150 @@
+from enum import Enum
```
janvonrickenbach, line 1: Moved from Simulation

src/ansys/dpf/post/simulation.py
```diff
@@ -720,78 +539,6 @@ def _create_dataframe(
         # Return the result wrapped in a DPF_Dataframe
         return df
 
-    @staticmethod
-    def _treat_cyclic(expand_cyclic, phase_angle_cyclic, result_wf):
```
janvonrickenbach, line 724: moved to _connect_workflow_inputs

src/ansys/dpf/post/simulation.py
```diff
@@ -979,74 +728,3 @@ def _build_selection(
                 time_freq_sets=[self.time_freq_support.n_sets]
             )
         return selection
-
-    def _requires_manual_averaging(
```
janvonrickenbach, line 983: moved to _build_workflow

src/ansys/dpf/post/result_workflows/_sub_workflows.py
```diff
@@ -0,0 +1,225 @@
+from typing import Any, Union
```
janvonrickenbach, line 1: Contains function to create all sub-workflows

src/ansys/dpf/post/result_workflows/_sub_workflows.py
```diff
@@ -0,0 +1,232 @@
...
+        [input_data_fwd, averaged_data_fwd, mesh_averaging_input_fwd]
+    )
+    mesh_averaging_input_fwd.connect(0, input_data_fwd, 0)
+
+    average_wf.set_input_name(_WfNames.input_data, input_data_fwd)
+    average_wf.set_output_name(_WfNames.output_data, averaged_data_fwd, 0)
+
+    skin_mesh_fwd = create_operator_callable(name="forward")
+
+    map_to_skin = True
```
janvonrickenbach, line 34: This has changed slightly. Before, the skin mapping never happened if force_elemental_nodal was false. This works for nodal data (where the skin mesh scoping only includes the skin nodes), but I believe it would fail for elemental data. In the new implementation, the mapping operator is always called. I had to add a server version check because the mapping operator did not support nodal data in older versions. (Fixed in this PR https://tfs.ansys.com:8443/tfs/ANSYS_Development/DPF/_git/DPF/pullrequest/450530)

tests/conftest.py
```diff
@@ -211,6 +211,15 @@ def grpc_server():
     server.shutdown()
```
janvonrickenbach, line 213: Automatically add license context to all tests, to minimize runtime.

PR #690, opened 2024-09-03
Average per body

janvonrickenbach:
Adds an averaging_config to the _get_results and _get_results_workflow functions of

- static_mechanical_simulation
- transient_mechanical_simulation
- modal_mechanical_simulation
- harmonic_mechanical_simulation

By default the averaging happens across bodies, which is the current behavior. If average_per_body is activated, the results are averaged per body and the resulting Dataframe/FieldsContainer contains separate fields for each body (with the a tuple of the selected properties with which a body is identified). The properties by which bodies are identified are configurable. The tests use the properties  "mat", "apdl_element_type", "elshape" and "apdl_real_id". The flag is currently not exposed for the results functions (stress, stress_nodal etc). The mesh support of the fields is always the full mesh.


src/ansys/dpf/post/fluid_simulation.py
```diff
@@ -254,6 +254,7 @@ def _get_result_workflow(
 
         _connect_initial_results_inputs(
             initial_result_workflow=initial_result_workflow,
+            split_by_body_workflow=None,
```
janvonrickenbach, line 257: not supported for fluid simulations

src/ansys/dpf/post/result_workflows/_build_workflow.py
```diff
@@ -101,8 +104,16 @@ def _requires_manual_averaging(
     category: ResultCategory,
     selection: Optional[Selection],
     create_operator_callable: Callable[[str], Operator],
+    average_across_bodies: bool,
 ):
     res = _result_properties[base_name] if base_name in _result_properties else None
+    native_location = res["location"] if res is not None else None
+
```
janvonrickenbach, line 111: We need to force elemental nodal if average_across_bodies is False

src/ansys/dpf/post/simulation.py
```diff
@@ -503,14 +503,17 @@ def _create_dataframe(
...
                 try:
-                    label_support = self.result_info.qualifier_label_support(label)
-                    names_field = label_support.string_field_support_by_property(
-                        "names"
-                    )
-                    values = [
-                        names_field.get_entity_data_by_id(value)[0] + f" ({value})"
-                        for value in values
-                    ]
+                    if label == "mat":
```
janvonrickenbach, line 506: I'm not sure this is the right way to do this. Using the code currently in the else block was not working for the "mat" label.

src/ansys/dpf/post/static_mechanical_simulation.py
```diff
@@ -66,14 +68,9 @@ def _get_result_workflow(
             mesh=self.mesh._meshed_region,
             location=location,
             force_elemental_nodal=result_workflows.force_elemental_nodal,
+            averaging_config=averaging_config,
         )
```
janvonrickenbach, line 73: Not needed anymore, the mesh is already available on the simulation object.

src/ansys/dpf/post/result_workflows/_utils.py
```diff
@@ -12,6 +13,29 @@ def __call__(self, name: str) -> Operator:
...
+    # List of properties that define a body. The mesh is split by these properties to
+    # get the bodies.
+    body_defining_properties: Optional[list[str]] = None
+    # If True, the results are averaged per body. The bodies are determined
+    # by the body_defining_properties.
+    average_per_body: bool = False
+
+
+default_per_body_averaging_config = AveragingConfig(
+    body_defining_properties=[
```
cbellot000, line 29: to prevent string duplications, could you please use ansys.dpf.core.common.elemental_properties?

src/ansys/dpf/post/result_workflows/_utils.py
```diff
@@ -12,6 +13,29 @@ def __call__(self, name: str) -> Operator:
...
+        "mat",
+        "apdl_element_type",
+        "elshape",
+        "apdl_real_id",
+    ],
```
cbellot000, line 34: By default do we split over all combinations of these properties?

src/ansys/dpf/post/result_workflows/_connect_workflow_inputs.py
```diff
@@ -79,14 +84,36 @@ def _connect_initial_results_inputs(
...
     The initial result workflow is the first workflow in the result workflows chain, which
     extracts the raw results from the data sources.
     """
+    selection_wf = selection.spatial_selection._selection
+
+    if selection.spatial_selection.requires_mesh:
+        selection_wf.connect(_WfNames.initial_mesh, mesh)
+
+    if averaging_config.average_per_body:
+        _enrich_mesh_with_property_fields(
```
cbellot000, line 100: I thought that the Operator splitting the mesh with a scoping had an option for that?

cbellot000 (approved):

src/ansys/dpf/post/result_workflows/_utils.py
janvonrickenbach: This configuration is currently only used in the tests. We are still evaluating what are the best properties to split so we get the bodies the user expects. Once we have consolidated that a bit more we can define a default.

src/ansys/dpf/post/result_workflows/_connect_workflow_inputs.py
janvonrickenbach: I don't see anything related in the docs: https://dpf.docs.pyansys.com/version/stable/operator_reference_load.html

janvonrickenbach: No its scoping::by_property. We split the scoping and average the results separately. But all the results have still the original mesh as a support.

src/ansys/dpf/post/result_workflows/_utils.py
```diff
@@ -12,6 +14,27 @@ def __call__(self, name: str) -> Operator:
...
+    body_defining_properties: Optional[list[str]] = None
+    # If True, the results are averaged per body. The bodies are determined
+    # by the body_defining_properties.
+    average_per_body: bool = False
+
+
+default_per_body_averaging_config = AveragingConfig(
+    body_defining_properties=[
+        elemental_properties.material,
+        "mapdl_element_type_id",
```
PProfizi, line 32: @janvonrickenbach is this one specific to your use-case or is it the same as `apdl_element_type`?

src/ansys/dpf/post/simulation.py
PProfizi: @janvonrickenbach here the logic is to expose names whenever available instead of IDs. But I think the logic in the else is dedicated to fluid qualifiers.. I think this is going to need a rework.
I do not think we have access to material names, so what you did is ok for me.
I think in your case the prrevious code was supposed to end-up in the except.

src/ansys/dpf/post/result_workflows/_utils.py
janvonrickenbach: As discussed: moved this default config which is not very generic to the tests

src/ansys/dpf/post/result_workflows/_utils.py
```diff
@@ -12,6 +14,27 @@ def __call__(self, name: str) -> Operator:
...
+
+    # List of properties that define a body. The mesh is split by these properties to
+    # get the bodies.
+    body_defining_properties: Optional[list[str]] = None
+    # If True, the results are averaged per body. The bodies are determined
+    # by the body_defining_properties.
+    average_per_body: bool = False
+
+
+default_per_body_averaging_config = AveragingConfig(
```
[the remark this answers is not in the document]
janvonrickenbach: As discussed: moved this default config which is not very generic to the tests

PR #739, opened 2024-10-29
Fix scoping if force_elemental_nodal is true.

janvonrickenbach:
When manual averaging is needed we request the data as ElementalNodal and perform the nodal averaging manually (force_elemental_nodal=True). If the mesh_scoping for a source operator (such as S) is nodal, and the requested location is ElementalNodal it returns the inclusive data (all elements that have at least one node in the scoping). This leads to one extra layer of data if
* the initial scope is nodal (see https://github.com/ansys/pydpf-post/issues/731) 
* The initial scope is elemental but the requested data is nodal. This happens because _build_selection converts the elemental scoping into a nodal scoping when the requested result is nodal.

In build selection, there was already a call to _requires_manual_averaging, based on which the location was modified. However, _requires_manual_averaging was always called with selection = None, which basically meant it always returns false and the location is not adapted. 

This PR adresses the problem by making the has_skin, and has_external_layer inputs explicit in the call to _requires_manual_averaging and passing the correct values in _build_selection. It also moves the call to _requires_manual_averaging outside of all the if statements so is called in the same way as in _get_results.

Notes: 

* This problem was not often visible because manual averaging is mostly needed in the skin case, where the final result is rescoped to the skin. It is more apparent with the new average per solid feature
* I removed the requires_manual_averaging method from selection and spatial selection, because it is incomplete. Some decisions depend on the result type (is it a strain result?). If these methods are needed for backward compatibility I can add calls to the new generic method with dummy arguments for the result types.

Also #fixes https://github.com/ansys/pydpf-post/issues/731


src/ansys/dpf/post/result_workflows/_build_workflow.py
```diff
@@ -110,12 +111,18 @@ def _requires_manual_averaging(
         return True
     if category == ResultCategory.equivalent and base_name[0] == "E":  # strain eqv
         return True
```
janvonrickenbach, line 113: Moved requires_manual_averaging in one place, because it is confusing if one part of the decision is done in the selection and the other part is done in the free function.

src/ansys/dpf/post/result_workflows/_build_workflow.py
```diff
@@ -96,7 +96,8 @@ def _requires_manual_averaging(
     base_name: str,
     location: str,
     category: ResultCategory,
```
janvonrickenbach, line 98: This was called with selection = None in the _build_selection, which meant the skin and external_layer did not have an effect.

src/ansys/dpf/post/selection.py
```diff
@@ -580,6 +580,36 @@ def select_nodes_of_elements(
             _WfNames.scoping, op.outputs.mesh_scoping_as_scoping
         )
```
janvonrickenbach, line 582: This is a new method. Before, requesting an elemental scoping with a nodal input yielded an error. If this was blocked intentionally, please let me know.

src/ansys/dpf/post/simulation.py
```diff
@@ -601,6 +601,26 @@ def _build_selection(
             return selection
         else:
             selection = Selection(server=self._model._server)
```
janvonrickenbach, line 603: This is the main change. Take it out of the if statements and takes into account the skin and external_layer flags passed to _build_selection.

src/ansys/dpf/post/simulation.py
```diff
@@ -658,11 +667,9 @@ def _build_selection(
                 selection.select_elements(elements=element_ids)
         elif node_ids is not None:
             if location != locations.nodal:
```
janvonrickenbach, line 669: Needed if requested if the input is a nodal selection, but requires_manual_averaging returns true.

src/ansys/dpf/post/selection.py
```diff
@@ -580,6 +580,36 @@ def select_nodes_of_elements(
...
 
+    def select_elements_of_nodes(
+        self,
+        nodes: Union[List[int], Scoping],
+        mesh: Mesh,
+    ) -> None:
+        """Select all elements of nodes using the nodes' IDs or an nodal mesh scoping.
```
PProfizi, line 588:
```suggestion
        """Select all elements of nodes using the nodes' IDs or a nodal mesh scoping.
```

src/ansys/dpf/post/simulation.py
```diff
@@ -601,6 +601,26 @@ def _build_selection(
             return selection
         else:
             selection = Selection(server=self._model._server)
+
+        if isinstance(skin, bool):
+            has_skin = skin
+        else:
+            has_skin = len(skin) > 0
```
PProfizi, line 608: @janvonrickenbach what is the use-case for `skin` to be a list of `int`?

tests/test_simulation.py
```diff
@@ -3810,10 +3989,23 @@ def test_averaging_per_body_nodal(
             mat_ids=bodies_in_selection,
             components=components,
             skin_mesh=res._fc[0].meshed_region,
+            additional_scoping=additional_scoping,
         )
     else:
-        # get reference data from mechanical
-        ref_data = get_ref_per_body_results_mechanical(ref_files[result], mesh)
+        # Cannot take reference for Mechanical because the named selection
```
PProfizi, line 3995: @janvonrickenbach so what is the reference then?

janvonrickenbach:
@PProfizi @cbellot000 Thanks for the review. I did some more testing on this an realized the current approach does not work for all cases. For instance, if force_elemental_nodal is true because the result type is equivalent strain, a nodal selection with a single node does not work correctly. The resulting selection is empty, because the scoping transposition is exclusive. I think I'll need to use an inclusive selection in this case. To get the right selection in the end, I will keep the original selection and reapply it in the end of the work flow. I will also add the inclusive flag argument as suggested by Paul.

src/ansys/dpf/post/simulation.py
janvonrickenbach: It is an element selection. The skin is extracted for these elements. This was already part of the api.

tests/test_simulation.py
janvonrickenbach: The elemental nodal result is rescoped to the additional_scoping and then averaged. I added a comment.

src/ansys/dpf/post/selection.py
```diff
@@ -580,6 +580,36 @@ def select_nodes_of_elements(
...
+        """
+        if isinstance(nodes, Scoping):
+            scoping = nodes
+        else:
+            scoping = Scoping(location=locations.nodal, ids=nodes, server=self._server)
+
+        op = operators.scoping.transpose(
+            mesh_scoping=scoping,
+            meshed_region=mesh._meshed_region,
+            inclusive=0,
```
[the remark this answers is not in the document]
janvonrickenbach: I added this as a parameter

src/ansys/dpf/post/simulation.py
```diff
@@ -649,7 +668,9 @@ def _build_selection(
                 )
         if named_selections:
             selection.select_named_selection(
-                named_selection=named_selections, location=location
+                named_selection=named_selections,
+                location=location,
+                inclusive=requires_manual_averaging,
```
janvonrickenbach, line 673: Inclusive in case of "requires manual averaging" to be consistent with a selection by a node_ids array. Not inclusive for the standard case to be backward compatible.

janvonrickenbach:
@cbellot000 @PProfizi I enhanced this PR: The selection is now inclusive for nodes (from a nodal named selections or a node id array), to support the case of nodal selections that don't select all the nodes of an element. Depending on the selection this results in an additional "layer" of nodes in the result. This is solved by returning the original selection from _build_selection and applying it in the end of the results workflow.

PR #745, opened 2024-11-08
Update for numpy 2

PProfizi:
Switch `MeshInfo.cell_zones` and `MeshInfo.face_zones` keys to `int` instead of `np.int32`.

Starting with `numpy==2`, the `np.int32` is reported explicitly as such, making their representation pop up where used directly. To circumvent that, when building maps based on `scoping.ids`, take the underlying Python `int`.

See [here](https://github.com/ansys/pydpf-core/actions/runs/11727755035/job/32669608017?pr=1847#step:17:1082) for the failing checks when upgrading to `numpy==2`.


BClappe (approved):
Ah ! Got a "gotcha"!

PR #757, opened 2024-11-18
Target 25.2 in ci

ansjhajjar:


PProfizi (changes requested):
You also need to add 251 to retro tests matrix

PR #763, opened 2024-11-25
Add test for single node selection

janvonrickenbach:


tests/test_simulation.py
```diff
@@ -3853,6 +3853,43 @@ def get_ref_per_body_results_skin(
 )
 
 
+def get_custom_scope(selection_name: str, mesh: MeshedRegion):
```
PProfizi, line 3856: Hi @janvonrickenbach I realize there are a lot of helper functions you made for tests. Do you think maybe some of them could be translated to helpers in the sources?

PProfizi (approved):

PR #779, opened 2024-12-18
Correctly pass shell layer argument.

janvonrickenbach:
Fix unit of sweeping phase operator


src/ansys/dpf/post/result_workflows/_sub_workflows.py
```diff
@@ -266,7 +266,7 @@ def _create_sweeping_phase_workflow(
             raise ValueError("Argument sweeping_phase must be a float.")
         sweeping_op = create_operator_callable(name="sweeping_phase_fc")
         sweeping_op.connect(2, sweeping_phase)
```
janvonrickenbach, line 268: Not sure why this only popped up now. Probably this workflow is not tested otherwise.

cbellot000 (approved):

PR #808, opened 2025-02-24
Use new Workflow.safe_connect_with

PProfizi:
Use `Workflow_safe_connect_with` when some inputs or outputs can be unavailable, as `Workflow.connect_with` now raises an error when an input or output is unavailable for connection.
Goes with https://github.com/ansys/pydpf-core/pull/2110


PProfizi:
New solution was chosen in PyDPF-Core's PR: added a `permissive` parameter to `Workflow.connect_with()` which defaults to True and filters input and output names to available combinations.

PR #823, opened 2025-04-17
fix bug 1250066 concerning point with skin extraction

MichaelNale:
change add_beam to add_beam_point to be compliant with changes server side

### DIFF
--- a/src/ansys/dpf/post/selection.py
+++ b/src/ansys/dpf/post/selection.py
@@ -447,7 +447,7 @@ class SpatialSelection:
         if self._server.meet_version("10.0"):
             # Add beam argument available since 9.1, but produces inconsistent
             # facets_to_elem mappings before 10.0
-            skin_operator.inputs.add_beam_point(True)
+            skin_operator.inputs.add_beam(True)
         self._selection.add_operator(skin_operator)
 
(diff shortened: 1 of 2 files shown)


MichaelNale:
@rlagha 
This PR needs to be merged first in order to properly merge the one in TFS without breaking Post. Once server side is ok I will update those tests regarding the new functionality of this API (add_beam --> add_beam_point so some num_entity need to be update as more elements are taken into account)

MichaelNale:
@PProfizi I don't understand why commenting a test make those checks fail. Locally everything's fine BTW.

rlagha (approved):

rafacanton:
@MichaelNale @PProfizi Because the failure in the examples is unrelated to this change. It is related to the issue in this PR: https://github.com/ansys/pydpf-post/pull/820

rafacanton (approved):

janvonrickenbach:
@MichaelNale Could you not add an _additional_ input "add_beam_point"  and keep the current add_beam input. You could then deprecate the add_beam input. There are other clients such as Result Explorer that will break due to this change.

PProfizi (changes requested):
See comment from @janvonrickenbach

PR #825, opened 2025-04-23
chore(test): temporarily skip tests during bug fix server-side on shell layers

MichaelNale:
Change a test post side to be compliant with changes server side


PProfizi (approved):
To revert later

PR #831, opened 2025-05-13
ci(doc): build using Python 3.11

PProfizi:
latest vtk dropped support for 3.10


PProfizi:
Let's first properly drop support for 3.9 and move the CI to 3.10 by default.

PR #839, opened 2025-05-23
ci(issues): Update issue templates for 26R1 dev

PProfizi:


.github/ISSUE_TEMPLATE/bug.yml
```diff
@@ -83,14 +86,15 @@ body:
...
-       - '3.9'
-       - '3.10'
-       - '3.11'
+       - '3.13'
        - '3.12'
+       - '3.11'
+       - '3.10'
+       - '3.9'
+       - '3.8'
+       - '3.7'
```
moe-ad, line 97: We no longer support 3.9 and below right?

PProfizi: For older versions of DPF yes we do. The issues are not only for the latest version of the client/server.

PR #840, opened 2025-05-23
261 versioning

guptadivya1501:


PProfizi:
Hi @guptadivya1501, thanks for this but I am making it use a repo variable in this PR https://github.com/ansys/pydpf-post/pull/838

PProfizi:
Hi @guptadivya1501, thanks for that. What I did though is make use of the repository variable here. https://github.com/ansys/pydpf-post/pull/838

PR #852, opened 2025-06-12
Reuse model instance when creating a simulation

FedericoNegri:


src/ansys/dpf/post/fluid_simulation.py
```diff
@@ -51,6 +51,46 @@
 from ansys.dpf.post.species import SpeciesDict
 
 
+def _create_dpf_model(
```
rafacanton, line 54: What do we gain by moving this here?

FedericoNegri: It just makes the constructor logic a bit more readable, avoiding 3 levels of nested if conditions.

PProfizi (approved):

PProfizi:
Thanks @FedericoNegri !

PR #864, opened 2025-07-08
fix(plotting): Fix docstrings tests as DPFPlotter now returns the plo…

PProfizi:
…tter instance

### DIFF
--- a/src/ansys/dpf/post/mesh.py
+++ b/src/ansys/dpf/post/mesh.py
@@ -499,7 +499,6 @@ class Mesh:
         >>> simulation = post.StaticMechanicalSimulation(example_path)
         >>> mesh = simulation.mesh
         >>> mesh.plot()
-
         """
         return self._core_object.plot(**kwargs)
 


src/ansys/dpf/post/mesh.py
```diff
@@ -499,7 +499,6 @@ def plot(self, **kwargs):
         >>> simulation = post.StaticMechanicalSimulation(example_path)
         >>> mesh = simulation.mesh
         >>> mesh.plot()
```
PProfizi, line 501:
```suggestion
        >>> mesh.plot()
        ...
```

src/ansys/dpf/post/meshes.py
```diff
@@ -182,6 +182,7 @@ def plot(self, **kwargs):
         ...    properties=[elemental_properties.material, elemental_properties.element_shape]
         ... )
         >>> meshes.plot()
+        ...
```
PProfizi, line 185:
```suggestion
        (...)
```

Issue #883, opened 2025-08-08
Test failures for 241, 242, 251 on flprj fluid tests on Linux


PProfizi:
### Before submitting the issue

- [x] I have checked for [Compatibility](https://dpf.docs.pyansys.com/version/stable/getting_started/index.html#compatibility) issues
- [x] I have searched among the existing issues
- [x] I am using a Python virtual environment

### Description of the bug

https://github.com/ansys/pydpf-post/actions/runs/16825947611/job/47662305381#step:7:156

### Steps To Reproduce

https://github.com/ansys/pydpf-post/actions/runs/16825947611/job/47662305381#step:7:156

### Which Operating System are you using?

Linux

### Which DPF/Ansys version are you using?

Ansys 2024 R1, Ansys 2024 R2, Ansys 2025 R1

### Which Python version are you using?

3.10

### Installed packages

https://github.com/ansys/pydpf-post/actions/runs/16825947611/job/47662305381#step:7:156

PR #884, opened 2025-08-08
Skip failing fluid test on Linux

PProfizi:
See #883


tests/test_fluid_simulation.py
```diff
@@ -68,7 +71,10 @@ def cfx_simulation(self):
     def test_simulation_str(self, fluent_simulation):
         assert fluent_simulation is not None
         assert str(fluent_simulation)
-
+    @pytest.mark.skipif(
```
PProfizi, line 74:
```suggestion

    @pytest.mark.skipif(
```

PR #893, opened 2025-08-13
fix: proper LICENSE metadata for package

RobPasMue:
Currently, the license is not showing up properly for ansys-dpf-post. This is caused by the way in which the license is referred in the pyproject.toml file. It should be done in this way moving forward


RobPasMue:
We can wait for now if you prefer - but please let me know once there is a new release. Thanks @moe-ad

moe-ad:
Noted. Thanks once again @RobPasMue.

PR #898, opened 2025-08-21
Do not turn element scoping into a nodal one

FedericoNegri:


tests/test_simulation.py
```diff
@@ -4694,3 +4685,51 @@ def test_beam_results_on_skin(beam_example):
     assert element_count_dict[element_types.Line2.value] == 40
 
     assert converted_field.max().data[0] == pytest.approx(190, 1e-2)
+
+
+def test_nodal_averaging_on_elemental_scoping(average_per_body_two_cubes):
```
janvonrickenbach, line 4690: Maybe add a comment that reference is from mechanical

janvonrickenbach (approved):

src/ansys/dpf/post/simulation.py
```diff
@@ -659,26 +659,22 @@ def _build_selection(
         if requires_manual_averaging and location != locations.elemental_nodal:
             location = locations.elemental_nodal
 
+        result_native_location = None
+        if base_name in _result_properties:
+            result_native_location = _result_properties[base_name].get("location")
```
rafacanton, line 664: I'd prefer to avoid using the `_result_properties` dict if possible (https://github.com/ansys/pydpf-core/issues/2366). I know that it is currently being used in other places here, but this needs to be refactored. Could you work instead with the ResultInfo? Thanks!

PR #918, opened 2025-10-14
Remove usage of _result_properties dictionary

FedericoNegri:
Stop using the static `_result_properties` dict from PyDPF-Core and rather look up result location from the result info.

What's problematic: for operators not listed as available results, we fail to lookup the corresponding native location. That's the case for beam operators (like `B_N`, `B_M1`, etc.) that are sort of a wrapper around SMISC.

I've tried to make it explicit in `_get_native_location`, where I'm also throwing an exception if a result was not found, rather than ignoring it and then deliver wrong results (because averaging is misconfigured etc.).

### DIFF
--- a/src/ansys/dpf/post/result_workflows/_utils.py
+++ b/src/ansys/dpf/post/result_workflows/_utils.py
@@ -131,7 +131,7 @@ def _get_native_location(
     """Get the native location of a result from its base name."""
     res = next((r for r in available_results if r.operator_name == base_name), None)
 
-    # special case for beam results, which are extracted from NMISC
+    # special case for beam results, which are extracted from SMISC
     if res is None and base_name in [
         "B_N",
         "B_M1",
@@ -142,7 +142,7 @@ def _get_native_location(
         "B_T1",
         "B_T2",
     ]:
-        res = next((r for r in available_results if r.operator_name == "NMISC"), None)
+        res = next((r for r in available_results if r.operator_name == "SMISC"), None)
 
     if res is not None:
         return res.native_location


tests/test_simulation.py
```diff
@@ -2203,36 +2203,10 @@ def test_kinetic_energy(self, transient_simulation):
...
-        assert len(result._fc) == 1
-        assert result._fc.get_time_scoping().ids == [2]
-        field = result._fc[0]
-        op = transient_simulation._model.operator("BFE")
-        op.connect(9, post.locations.elemental)
-        field_ref = op.eval()[0]
-        assert field.component_count == 1
-        assert np.allclose(field.data, field_ref.data)
+        # the model does not contain structural temperature results
+        with pytest.raises(ValueError) as excinfo:
```
PProfizi: @FedericoNegri so this test used to silently fail at finding the native location, and returned something wrong?

src/ansys/dpf/post/result_workflows/_utils.py
```diff
@@ -122,3 +123,28 @@ def _append_workflow(new_wf: Optional[Workflow], last_wf: Workflow):
...
+        "B_N",
+        "B_M1",
+        "B_M2",
+        "B_MT",
+        "B_SN",
+        "B_EL",
+        "B_T1",
+        "B_T2",
+    ]:
+        res = next((r for r in available_results if r.operator_name == "SMISC"), None)
```
PProfizi: @FedericoNegri so all `SMISC` results have the same native location then?

tests/test_simulation.py
```diff
@@ -2203,36 +2203,10 @@ def test_kinetic_energy(self, transient_simulation):
...
+            _ = transient_simulation.structural_temperature(set_ids=[1])
+        assert "not found" in str(excinfo.value)
```
PProfizi:
```suggestion
        assert "Unknown native location" in str(excinfo.value)
```

src/ansys/dpf/post/result_workflows/_utils.py
```diff
@@ -122,3 +123,28 @@ def _append_workflow(new_wf: Optional[Workflow], last_wf: Workflow):
...
+
+    if res is not None:
+        return res.native_location
+
+    raise ValueError(f"Result with base name '{base_name}' not found.")
```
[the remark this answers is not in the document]
FedericoNegri: > This function checks for the native location, not whether the result exists.

I disagree, the function throws if it can't find the result, which is a prerequisite for querying its location.
_Unknown native location_ would be an appropriate error message for a native location set to None.

> we move the check for existing result with base name somewhere above

I've been thinking about this and considered having a function

```python
def _get_available_result(
    available_results: list[AvailableResult], base_name: str
) -> AvailableResult:
    res = next((r for r in available_results if r.operator_name == base_name), None)

    # special case for beam results, which are extracted from SMISC
    if res is None and base_name in [
        "B_N",
        "B_M1",
        "B_M2",
        "B_MT",
        "B_SN",
        "B_EL",
        "B_T1",
        "B_T2",
    ]:
        res = next((r for r in available_results if r.operator_name == "SMISC"), None)

    if res is None:
        raise ValueError(f"Result with base name '{base_name}' is not available.")

    return res
```

but I think it would just be misleading. You come in with `B_N` (axial force) and it returns you an available result with physical name 'Elemental Summable Miscellaneous Data', no unit etc.

The proper fix is to have those beam results be listed as available results (that was discussed internally).

PR #924, opened 2025-11-12
Handle native location query for NAR operators

FedericoNegri:
Workaround to the fact that NAR operators are exposed as `mapdl::rst::<operator-name>` (e.g. `mapdl::rst::NS`) but the operator name in the result info only includes `<operator-name>` (e.g. `NS`). These changes allow to query the native location of say `mapdl::rst::NS`.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -225,6 +225,13 @@ def beam_example():
     )
 
 
+@pytest.fixture()
+def nar_example():
+    return _download_file(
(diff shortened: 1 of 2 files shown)


PProfizi:
Hi @FedericoNegri just so I understand, why do we need to deal with `mapdl::rst::NS` in a specific way? Why not go through `NS`? When is `mapdl::rst::NS` called explicitly? Is it a workaround for your project?

PR #1007, opened 2026-05-29
ci: Fix beam element types after server update

rafacanton:
Adapt to changes in https://github.com/ansys/pydpf-core/pull/3235


PProfizi (approved):

rafacanton:
@PProfizi @moe-ad Not sure what's wrong with the Doc stage, it keeps failing just after it starts

PProfizi:
> @PProfizi @moe-ad Not sure what's wrong with the Doc stage, it keeps failing just after it starts

I am not sure, let's bypass.

PR #1032, opened 2026-09-09
Add `read_beams` option to internal workflow creation

FedericoNegri:
In 2027 R1, the default of the `read_beams` input for MAPDL operators has changed from False to True. In Result Explorer we don't want to get stress/strain/etc beam results, so we need a way to set this input on the result operator.

Exposing it on the internal `MechanicalSimulation._get_result_workflow` functions (which are the Result Explorer entry points) as an optional input.

No changes to the pydpf-post API or behavior.


rafacanton (approved):

janvonrickenbach (approved):
Looks good overall. Looks like some older server versions produce different "locations".

FedericoNegri:
> Looks good overall. Looks like some older server versions produce different "locations".

Updated the test to fix those failures.

